PR #2054, opened 2026-07-04
docs: task the daily assistant with driving Kyverno policy violations to zero

devantler:
## Why
Policy Reporter now gives a whole-cluster view of the Kyverno `PolicyReport`s the cluster already produces (see devantler-tech/platform#2459). The autonomous assistant should *act* on that — not just observe — so cluster compliance actually reaches and stays at zero violations.

## What
Adds a **"Policy compliance (Kyverno) — drive violations to zero and hold"** program to the platform product card, and makes the assistant aware of Policy Reporter as the surface for it. It mirrors the existing Kubescape *"drive to 100% and hold"* section: enumerate failing results each run (existing **and** newly introduced), root-cause-fix by default, graduate a cleared policy from Audit to Enforce, and add a scoped exception only when a finding is genuinely by-design.

## Notes
- Definition/self-improvement change — one product card, no behaviour outside the assistant's own instructions.
- Pairs with platform#2459 (which deploys Policy Reporter). Draft until you promote.

### DIFF
--- a/.claude/skills/products/platform/SKILL.md
+++ b/.claude/skills/products/platform/SKILL.md
@@ -1,6 +1,6 @@
 ---
 name: maintain-platform
-description: Repo + live-platform health/maintenance menu for devantler-tech/platform (GitOps — Kustomize overlays + Flux CD, Cilium/Talos/KSail/SOPS). Repo side — triage, manifest/Helm/Flux investigation & fixes, Helm chart + Actions version bumps, manifest cleanup, stale-PR nudges (static validation; never spin up a cluster to test a diff). Live side — read-only health investigation of the running prod cluster + its observability stack (Flux Kustomizations/HelmReleases, Coroot, Kubescape, Kyverno, Kubernetes events). Use when the daily maintainer selects Platform.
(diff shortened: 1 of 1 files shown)


devantler:
> 🤖 Generated by the Daily AI Assistant

Addressed CodeRabbit's `warn`-vs-`fail` point (commit d328d51): the platform SKILL card now names the **`fail` count as the hard zero-and-hold target** and tracks **`warn` in a separate tally** as a softer lead signal (an `Audit` policy heading to `Enforce`, or a deprecation notice) — triaged and worked down, but not sharing `fail`'s zero-gate. Thread auto-resolved on the edit. Back to review-ready.

devantler:
> 🤖 Generated by the Daily AI Assistant

**Addressed CodeRabbit's outside-diff-range finding** (`SKILL.md` 43-49, 🟡 Minor — scope policy inventory to all namespaces).

Valid — `Policy` and `PolicyReport` are namespaced, so the survey commands could miss objects outside the current namespace. Fixed in `98f36d6`: `get cpol,pol -A` and `polr,cpolr -A` (the `-A` is a no-op for the cluster-scoped `cpol`/`cpolr`, correct for the namespaced `pol`/`polr`).

PR #2109, opened 2026-07-10
docs: reconcile agent name to Daily AI Engineer + prefix-match disclosure disambiguator

devantler:

**Why:** The untrusted-input safety disambiguator — the guard that tells the agent's *own* disclosed comments apart from the human maintainer's instructions (a self-instruction-loop guard) — keyed on the exact string `> 🤖 Generated by the Daily AI Assistant`. But the actor was renamed to "Daily AI **Engineer**" everywhere else (heading, mandate, scheduled task, memory), and every recent comment posts the "Engineer" form. A strict exact-string read misclassifies the agent's own comments as maintainer instructions — exactly the failure the guard exists to prevent.

**What:** Matches the disambiguator (in AGENTS.md and both surveyor agents) on the stable `> 🤖 Generated by the Daily AI` **prefix** — recognising both the current "Engineer" and historical "Assistant" wording (the human maintainer posts neither) — and standardizes the actor name to "Daily AI Engineer" across the agent, surveyors, and skill descriptions so the definition no longer drifts. Markdown-only; safety-tightening (never loosens a guardrail). The historical ADR-0001 disclosure record is left intact (the prefix-match covers it).

Evidence: surveyor 665th flagged the drift live.

Part of the self-improvement loop.


devantler:

@coderabbitai Good catch — fixed in `5dbf9c9`: the self-improvement skill's Conventional Commit example now reads `chore(ai-engineer): …`, consistent with the renamed actor.

PR #2127, opened 2026-07-11
docs(constitution): routine prompts and runtime guards are monitored parts of the deployment

devantler:

## Why

You directed (2026-07-11) that each agent instance monitor and enhance its routine/schedule prompt, and — same direction, second half — its runtime permission layer (Claude classifiers / ChatGPT-side guards), keeping both current with the least privilege that still lets the mandate run effectively. The constitution must carry it so the sister ChatGPT/Codex agent inherits both duties from the shared brain.

## What

Adds two stewardship rules to the Self-improvement section plus a matching every-run check in the run-loop skill: (1) **Routine-prompt stewardship** — each instance sanity-checks its own dispatch prompt every run, fixes it directly with before/after recorded in memory and the report, propagates anything substantive into the version-controlled definition, and only ever tightens its backstop non-negotiables; (2) **Runtime guard/permission stewardship** — over-broad grants are tightened directly (recorded; sensitive specifics stay in the private host-audit notes), needed widenings are surfaced to you as one-clicks (an instance never self-widens), and the full review folds into the ~monthly host least-privilege audit. Neither instance touches the other's prompt or guards.

Applies to both instances as soon as this merges (both boot from this file).

### DIFF
--- a/.claude/skills/portfolio-maintenance/SKILL.md
+++ b/.claude/skills/portfolio-maintenance/SKILL.md
@@ -349,7 +349,15 @@ For each selected product:
 ## 5. Reflect & improve (self-learning)
 At the end of every run, record operational **`learnings`** in native memory (`learnings.md`) — steps
 that failed / were flaky / slow / wasted effort, coverage gaps, stale or ambiguous instructions,
-security/reliability weaknesses in your own workflow. **~Weekly** (or sooner for a clear high-value / security /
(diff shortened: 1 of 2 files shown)


devantler:

@codex review

devantler:
@coderabbitai run pre-merge checks

> 🤖 Generated by the Daily AI Assistant

PR #2139, opened 2026-07-13
docs(constitution): scripting stack = bash or Go, never Python (portfolio-wide)

devantler:

## Why
You directed (2026-07-13, interactive) that the portfolio's scripting stack is a constitutional decision: no Python anywhere — Go is the favorite, bash is a fine starting point that migrates to Go as scripts grow. Only the platform repo had this rule until now.

## What
Adds a *Scripting stack* section to the shared engineering contract: never introduce Python in any repo, bash-first-then-Go as the standard maturation path, and existing Python becomes an oldest-first migration target (inventory sweep tracked in #2140).


devantler:

@codex review

PR #2207, opened 2026-07-17
docs(agents): World at Ruin is a first-class product, not lowest priority

devantler:

## Why

Your direction today: World at Ruin was not meant to be less prioritized — it needs the same attention and love as the rest of the portfolio. The bootstrap-day definition baked "lowest priority / pick it up only when nothing else demands attention" into three places, which would keep every future run treating it as back-of-queue.

## What

Rewords the constitution's portfolio-map note, the run-loop's card index, and the product card: World at Ruin is a first-class product in the normal selection and fairness rotation. The game repo's own AGENTS.md gets a matching PR.

Supersedes the 2026-07-16 lowest-priority note.


devantler:

@codex review

PR #2210, opened 2026-07-17
test(worktrees): add submodule-init.sh regression guard + CI wiring

### LINKED ISSUE #2174: test(worktrees): add a regression guard for submodule-init.sh

## Problem
`.claude/scripts/submodule-init.sh` is a **fail-closed safety tool** — agents trust its verdict before editing a submodule worktree, and a wrong "all isolated" verdict silently collapses parallel sessions into one physical tree. It ships with **no test harness**, while its sibling primitive `.claude/scripts/safe-clone.sh` has `safe-clone.test.sh`.

That gap is not theoretical. During review of #2164 the reviewer found **four** real defects in three rounds, including a **fail-open**: when a stray `core.worktree` pointed at another *valid* checkout — i.e. exactly the collision the tool exists to detect — the submodule was silently dropped from the sweep and the script exited 0, "all isolated". Each round was verified only by a hand-run probe, and one round's manual verification passed only because it happened to avoid the broken case.

## Proposed direction
Add `.claude/scripts/submodule-init.test.sh`, modelled on `safe-clone.test.sh`, exercising the tool against throwaway fixture repos:
- clean tree → exit 0, every submodule reported isolated;
- stray `core.worktree` pointing at another valid checkout → **exit non-zero**, names the colliding checkout (the fail-open regression);
- deinitialised submodule → skipped, no false alarm;
- repair path clears the stray key and the re-probe passes;
- run from both the main checkout **and** a linked worktree (where the submodule gitdir lives under `.git/worktrees/<wt>/modules/<path>`).

Wire it into CI alongside the existing script guard so the next regression fails a check instead of a production run.

## Rough size
Small–medium (one test script + CI wiring).

Part of #2164.

devantler:

**Why.** `submodule-init.sh` is a fail-closed safety tool — agents trust its "isolated" verdict before editing a submodule worktree, and a wrong verdict silently collapses parallel sessions into one physical tree. It shipped with **no test harness** (its sibling `safe-clone.sh` has one), and the #2164 review found four real defects in three rounds — including a fail-open where a stray `core.worktree` pointing at another valid checkout was silently dropped and the script exited 0 "all isolated".

**What.** Adds `.claude/scripts/submodule-init.test.sh` (modelled on `safe-clone.test.sh`, hermetic file:// submodule fixtures, no network) covering the clean tree, the stray-`core.worktree` collision (must fail and name the collider — the fail-open regression), repair, a deinitialised submodule (skipped, no false alarm), and the linked-worktree execution model. Wired into CI on ubuntu + macOS beside the existing script self-tests. RED/GREEN proven locally (reintroducing the fail-open fails the collision assertions).

Fixes #2174
Part of #2164

### DIFF
--- a/.claude/scripts/submodule-init.test.sh
+++ b/.claude/scripts/submodule-init.test.sh
@@ -1,0 +1,149 @@
+#!/usr/bin/env bash
+#
+# Self-test for submodule-init.sh — proves the worktree-isolation guard DETECTS
+# a stray core.worktree that collapses a submodule onto another checkout (the
+# fail-open the #2164 review caught across three rounds), that a clean tree
+# passes, that repair clears the stray key and pins it per-worktree, that a
+# deinitialised submodule is skipped without a false alarm, and that all of this
+# holds when the tool is invoked from a linked superproject worktree (the
+# documented agent execution model, where the submodule gitdir lives under
+# .git/worktrees/<wt>/modules/<path>).
+#
+# Fixtures are throwaway local `git init` repos wired as file:// submodules — no
+# network, no real submodules touched. Run in CI so a refactor that re-opens the
+# fail-open is caught here, not by a silent cross-session collision in a
+# production run (the exact papercut submodule-init.sh exists to prevent).
+set -Eeuo pipefail
+
+here="$(cd "$(dirname "${BASH_SOURCE[0]}")" && pwd)"
+helper="$here/submodule-init.sh"
+tmp="$(mktemp -d)"
+trap 'rm -rf "$tmp"' EXIT
+
+# Hermetic git environment: the host's real system/global config must not affect
+# fixtures, and file:// submodules need protocol.file.allow (default-denied since
+# the CVE-2022-39253 hardening) — set once in the throwaway global config so the
+# script's own internal `git submodule update --init` inherits it too.
+export GIT_CONFIG_NOSYSTEM=1
+export GIT_CONFIG_GLOBAL="$tmp/gitconfig"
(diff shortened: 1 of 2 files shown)


devantler:

**Tried-and-evaluated-as-a-user (readiness record).** I exercised the guard, not just reasoned about it:

- **GREEN** — ran `bash .claude/scripts/submodule-init.test.sh` against the real script: all 15 assertions pass (`submodule-init self-test: all cases passed`). Each scenario was first prototyped against real `git submodule add` / `git worktree add` fixtures to pin actual behaviour (relative self-referencing `core.worktree` on a clean add; absolute stray → `show-toplevel` resolves to the collider; repair moves the key into `config.worktree`; linked-worktree gitdir under `.git/worktrees/<wt>/modules/sub`).
- **RED** — reintroduced the historical fail-open (made `resolves_to_itself` return 0) in a throwaway copy; the broken script reports `sub — isolated ✓` on a colliding submodule and the three collision assertions FAIL, exiting non-zero. So the guard genuinely catches the regression it exists for.
- Lint: `shellcheck -s bash` clean (exit 0); `actionlint` clean on the ci.yaml wiring; `yq` confirms the new `test-submodule-init` job is present in the graph.

Requesting a review once CI settles green; will self-promote only on the full pentad.

devantler:

@codex review

Test-only + CI-wiring change at head `42ef62d0`. `test-submodule-init` passes on both ubuntu and macOS in CI; required checks green, mergeState CLEAN. Requesting a review at head for the green-review gate.

PR #2211, opened 2026-07-17
docs(agents): sync World at Ruin product card to the refined product law

devantler:

**Why.** You refined World at Ruin's product law (2026-07-17): the world *can* be migrated, so the absolute "no undo" framing is wrong — what matters is that a player is never surprised. The product card duplicated the old wording and would otherwise drift from the authoritative law.

**What.** Syncs the card's Product-law section to match [world-at-ruin#63](https://github.com/devantler-tech/world-at-ruin/pull/63): migration allowed when non-breaking or via a player-visible deprecation; experimental features opt-in behind a default-off flag; anti-inflation constraint kept without the "no undo" wording.

Definition change (`.claude/` card) — **keeps the human promotion gate**; promote when you're happy and I'll drive it to merge. Land alongside or after world-at-ruin#63.


devantler:

@codex review

PR #2228, opened 2026-07-18
docs: make site docs read for humans, not machines

devantler:
## Why

Several pages on the site read like an inventory rather than an explanation. The platform template's "What's Inside" listed around 40 product names with no hint of what any of them actually do for you — useful only to someone who already knew the whole stack. The about page repeated the same "Developing and maintaining X" bullet six times across roles.

## What

Rewrote the densest pages so a reader learns what they get. Product names stay — someone looking for a Talos-based platform still needs to see "Talos" — but each group now says what it gives you. Also cut the repetition on the about page and fixed a product-name typo.

No behaviour or configuration changes; docs prose only. The CI drift-check markers are untouched.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #2230, opened 2026-07-18
docs(agents): require jargon-free voice for user-facing documents

devantler:
## Why

Our public docs had drifted into writing that reads like an inventory rather than an explanation — the platform template listed around 40 product names with no hint of what any of them do for a reader, and the about page repeated the same bullet six times. Nothing in the contract said docs had to be written for a human, so nothing caught it.

## What

Makes the jargon-free voice a standing requirement for every user-facing document, recorded in both places that carry docs conventions (`AGENTS.md` and the product-engineering skill).

The rule carries its own calibration so it can't be misread as a blanket ban on technical terms: for technical readers the product names stay, since removing the words that let someone identify what they're getting is a regression, not simplification. The strict version applies only to genuinely non-technical readers.

Definition change — leaving it to you to promote. Prose-only; no behaviour or configuration changes.

Companion docs PR: #2228


devantler:
> 🤖 Generated with Claude Code (interactive session)

@codex review

PR #2234, opened 2026-07-18
docs(agents): add the parallel-session claim protocol to the shared contract

### LINKED ISSUE #102: Parallel-session claim protocol: six end-to-end builds were discarded in one day because a PR is the only claim signal

## Evidence

World at Ruin is built by **many agent sessions in parallel**, and on 2026-07-17/18 that produced repeated **fully-duplicated builds** — two sessions independently implementing the same issue, one of them discarding complete, validated work at push time:

| Issue | Outcome |
|---|---|
| #66 swept collision | **Two** separate sessions each built it end-to-end (sweep + tests, RED/GREEN, golden-verified) before discovering PR #74 already owned it. Both discarded. |
| #81 foliage scatter | A session built the full library + committed golden + 5 negative controls; at push time found PR #83 had merged ~8 minutes earlier. Discarded. |
| #88 telegraph resolution | A session pushed a claiming branch at 06:41:26Z and built the full resolver + tests; PR #94 was opened at 06:42:18Z — **52 seconds later** — by another session. Discarded. |
| #86 armour axis | Claimed by PR #93 within minutes of another session evaluating it as its next pick. |

That is **at least four issues** and **six discarded end-to-end builds** in roughly one day. The work was correct; only the *coordination* failed.

## Audience / problem

The waste is invisible in the repo — merged PRs look fine — but it is large and systematic, and it falls hardest on the most substantive picks (a session that takes a *big* issue spends the longest building and is therefore the most likely to be beaten to the PR).

The root cause is that **an open PR is the only recognised claim**, and a PR only appears at the *end* of a build. So the claim signal arrives exactly when it is too late to prevent the collision. Two sessions surveying the same backlog minutes apart will reliably pick the same issue — they are running the same oldest-actionable-first selection rule over the same data, so convergence is the *expected* behaviour, not bad luck.

## Hypothesis

Moving the claim to the **start** of the build, and making the pre-build check read every claim signal rather than only PRs, collapses the collision window from "the whole build" to "the few seconds between check and claim".

## Smallest useful change

A documented claim protocol in `AGENTS.md`, costing one command before building:

1. **Claim before building.** Once an issue is selected, immediately (a) self-assign it and (b) push an empty `claude/war-<slug>` branch. Both are cheap, visible, and reversible.
2. **Check all three signals before selecting**, not just open PRs: open PRs, remote `claude/*` branches, and issue assignees.
3. **Claims expire.** A claim with no PR after ~2 hours is stale and may be taken over — so an abandoned session never parks an issue permanently.
4. **Re-verify immediately before the first push** (already practised; keep it as the backstop for the residual seconds-wide window).

## Success signal

Zero fully-duplicated builds across a comparable multi-session day; any collision that does occur is caught at *selection* rather than at push. Measured by the absence of "discarded, sibling owned the lane" outcomes in run reports over the next week.

## Note for the maintainer — this touches the monorepo constitution

The shared contract currently states that **"a bare assignee does *not* reserve an issue — only an open PR does"**, which exists so a stale assignment cannot park work forever. Point 1 above partially conflicts with it, and point 3 is the reconciliation: a *time-boxed* claim reserves briefly, then decays. Two ways to land it:

- **repo-local** (this repo's `AGENTS.md` only), since World at Ruin is the only product running this many concurrent sessions; or
- **portfolio-wide**, as a constitution amendment — the same collision applies anywhere parallel sessions share a backlog.

Recommend repo-local first, promoting it if it proves out. Flagging rather than deciding, since it edits agent-definition surface.

## Rough size

Small — a documented convention, no code.

devantler:

## Why

Several agent instances work the same backlog at once, and they all follow the same rule: take the oldest actionable issue. So two sessions starting minutes apart reliably pick the *same* one. The only signal that an issue is taken is an open pull request — which appears when the work is already finished. By then the duplication has happened.

On World at Ruin that cost **six completed, validated pieces of work thrown away in about a day**. Two of them were lost by 52 and 135 seconds.

## What

Moves the "I'm taking this" signal to the **start** of the work instead of the end: claim the issue and push the branch as soon as it is picked, then do the work. Claims lapse after a couple of hours so nothing stays parked by a session that died. On a lost race, the losing side now has a defined use — comparing the two attempts reliably finds real bugs in the one that shipped.

Also fixes a specific way the current check fails: matching on branch names lets British and American spellings of the same word hide two sessions from each other. Matching on the issue number does not.

Kept deliberately narrow — it changes *when* work is claimed, not *what* work gets picked, and it does not relax any safety rule.

Fixes devantler-tech/world-at-ruin#102

> **Definition change — promotion is yours.** Per *Self-improvement*, definition PRs keep the human gate; I will not self-promote this. Evidence is from my own runs only.


devantler:

This is a change to the engineering contract itself, so the useful review is consistency rather than code: does the new *Claim protocol* section contradict any existing rule, does the reconciliation of the "a bare assignee does not reserve an issue" rule actually preserve that rule’s intent (a stale assignment must never park work permanently), and does anything here weaken a safety guardrail rather than tighten it. The two skills are edited only to stop them restating the superseded wording.

@codex review for internal consistency with the rest of AGENTS.md and for any weakened guardrail

devantler:

All five findings were valid and are fixed in `4d054d9`. Two were guardrail-adjacent and I'm glad they were caught — a contract that quietly opens a hole is worse than one that never mentioned the topic.

**P1 — limit claims to in-scope repos.** Correct. The section said *"on every repo"* and then instructed self-assigning and pushing a branch. **Claiming is a write action**, so that phrasing could authorise a first touch of a repo the *Professional-work repository boundary* has not cleared — and that boundary forbids even read-only probing. Now scoped to in-scope `devantler-tech` repos, with the boundary named as winning outright and an explicit "nothing here licenses a first touch of an unconfirmed repo".

**P1 — do not execute external winners.** Also correct, and the sharper of the two: *"post only findings you have executed against their code"* would have an agent check out and run an **external-contributor** branch, which the trust gate forbids outright. Split by author — execute the probe against a **trusted/routine-owned** winner, **static review only** against an external one, and state in the finding which it was. The same sentence was repeated in `product-engineering`, so it is fixed there too.

**P2 — claim branches discoverable by issue.** Right: before a PR exists there is no body to grep, so `claude/<area>-<desc>` left a rival only the normalised-stem match that #96 showed is fragile. Branch names now carry the issue number (`claude/<area>-<desc>-<issue>`) — the one token that cannot be spelled two ways.

**P2 — define the claim clock.** Right, and it would have misfired in the common direction: a claim branch usually points at the **base commit**, so its date reads as long expired and every fresh claim would look stale. Timed from the issue's `assigned` timeline event instead, with the exact `gh api …/timeline` query given, plus the two degenerate cases (assigned-without-branch, branched-without-assignment) defined as no claim.

**P2 — live claims in the skip rules.** Right, and a genuine contradiction: a freshly claimed oldest issue was both un-takeable and un-skippable. Added as skip reason **(e)** in *Drain oldest-first* itself — not just referenced from here — and flagged as the only skip reason that expires on its own, so it cannot become a permanent park.

**One I found myself while waiting**, also in this push: the protocol tells agents to read assignees as a claim signal, but every instance assigns as `devantler` — so an agent could read that login as the *human maintainer* having taken the issue and skip it, parking the work instead of claiming it. Clarified that an assignee means *an instance* claimed it, never a hands-off signal.

Also merged `main` in — #2220 (agent self-review fallback) and #2230 (jargon-free voice) had landed. Verified after merging that both of their sections survived intact and my net diff is still only these three files.

Re-requesting, since the push stales the previous result.

@codex review

devantler:

All six fixed in `d1cc54c`. Two of these showed the protocol **would not actually have worked**, which is the most useful kind of finding on a contract change — it reads fine and fails in practice.

**Surveyor blindness (the big one).** In the scheduled flow, issue selection runs off the `portfolio-surveyor` digest — and that agent fetched issues with `number,repository,title,labels,updatedAt,url` and reported neither assignees nor `claude/*` branches. So the claim signals I was telling agents to check were **structurally invisible on the primary path**, and the duplicate-build race stayed wide open in exactly the flow the protocol exists for. The surveyor now returns `assignees`, reports `claude/*-<issue>` branches on PR-less issues, and carries a `CLAIMED …` line in the digest.

**Stale-claim takeover never started a new lease.** Correct, and subtle: every instance assigns the same `devantler` login, and GitHub's add-assignees endpoint is a no-op for an already-assigned user — so a takeover produced **no new `assigned` event**, the lease kept the dead claim's timestamp, and the next run would read the fresh claim as already expired and race. Takeover now unassigns before re-assigning.

**Branch templates.** Right — I added the requirement in one place while the templates agents actually copy still said `claude/<area>-<desc>` in the execution model, `portfolio-maintenance`, and the monorepo card. A run following those produced an unmatchable claim branch, defeating the fix. All three aligned, and the HANDS-OFF identification rule updated so routine-vs-interactive recognition stays accurate against the new shape (older routine branches predate the number — noted).

**`product-engineering` skip list.** Right — it keeps its own `(a)–(d)`, so a run invoking that skill directly could neither take nor skip a claimed issue. Added `(e)` there.

**Timeline query.** Right on both counts: no `--paginate` and no selection, so a long-lived issue hands back an old assignment from page 1. Now `--paginate` with `|last`.

Worth saying plainly: I had claimed in my previous reply that the branch-number and skip-list fixes were done. They were done in the *contract* but not in the *copies*, which is precisely the drift the definition-of-done rule about updating every agent file in the same PR exists to catch. Fair catch.

Re-requesting at the new head.

@codex review

devantler:

Three new findings this round (the `10:24:09Z` ones — the earlier bodies re-surface in a head-anchored query because GitHub re-anchors review comments to the new head; rounds 1–2 are fixed and resolved). All three were real, and all three were **caused by my round-2 fixes** — worth stating plainly rather than presenting as fresh discoveries.

**Aggregate paginated pages — valid, but the suggested remedy does not work.** The diagnosis is exactly right: under `--paginate` each page is a separate JSON array, so `'[…]|last'` runs *per page* and returns the last match of the final page rather than the newest overall. I tried the proposed `--slurp` and `gh` rejects it: `the --slurp option is not supported with --jq`. So the fix emits one timestamp per line and takes the max in the shell instead:

```sh
gh api repos/<o>/<r>/issues/<n>/timeline --paginate \
  --jq '.[]|select(.event=="assigned")|.created_at' | sort | tail -1
```

Verified against a real multi-event timeline (w-a-r#109) — both this and the `--slurp | jq` pipe return the same newest timestamp; this form avoids the flag conflict.

**Qualify issue edits by repository — valid, and a live footgun here.** Runs work from the monorepo checkout or a submodule worktree, so an unqualified `gh issue edit <n>` resolves against *that* repo and would silently edit whatever issue happens to share the number. Both takeover commands now pass `-R <owner>/<repo>`.

**Don't label bare assignees as live claims — valid, and I introduced the contradiction.** My digest shape emitted `CLAIMED … claim-branch=none`, which would have made a bare assignee park an issue as skip reason (e) — exactly what *"a bare assignee does not reserve an issue"* forbids, and the failure mode the whole time-boxing design exists to avoid. `CLAIMED` now requires **both** an assignee and a matching claim branch; an assignment without a branch is reported as the ordinary open issue it is.

I also re-checked that the round-1/2 fixes are still intact after these edits (scoping, external static-review-only, branch templates in all three copies, `product-engineering` skip `(e)`, surveyor `assignees`) rather than assuming.

Re-requesting at the new head.

@codex review

devantler:

Field evidence from this tick (2026-07-18 ~10:2xZ), offered as data for this PR — not a request to change it.

I lost a lane to exactly the race this PR describes, and the details sharpen one edge of it.

I picked `wedding-app#176`. It read as **completely cold on every remote signal**: zero open PRs on the repo, ~13h since its last merge, and `git ls-remote` showed no matching branch. I only discovered the collision when `git worktree add` failed with `fatal: a branch named 'claude/wedding-lcp-render-delay-176' already exists` — a live sibling session already held a worktree and a commit on the **identical derived branch name**, created **one minute** before I looked.

Two things follow:

1. **The push-at-start rule is what fixes this** — the sibling had claimed locally but not pushed, so every GitHub-side check I could run was structurally blind. That is the gap this PR closes, and this is a clean example of it.
2. **During the transition, `git worktree list` on the submodule is a local oracle that sees claims GitHub cannot.** Until every instance is pushing at start, a sibling that hasn't adopted the protocol is invisible to `gh pr list` but visible in `git worktree list`. Might be worth a sentence as a belt-and-braces check.

Also a small corroboration of the branch-name point: this is now the **fourth** sighting of two instances independently deriving the *same* branch name from the same issue (WaR #81 and #86 previously). That makes the `worktree add` name collision a fairly reliable early stand-down signal in its own right — I treated it as one and dropped the lane rather than working around it with a suffix.

I stood down, pushed nothing, and took an uncontended lane instead ([agent-plugins#72](https://github.com/devantler-tech/agent-plugins/pull/72)).

devantler:

All four fixed in `ddbf9ac`. Two were genuine mechanism defects, two were narrower:

- **Bare-number search** — right, `--search "<issue> in:body"` matches any body containing those digits (a benchmark count, a date, another repo's number), which could hide the oldest actionable issue behind an unrelated PR. Anchored on the `#<issue>` reference with a concrete `gh pr list --search '"#<issue>" in:body'`.
- **Stale bare assignment** — right, and I'd only half-fixed it: the no-op assignee problem applies to *any* run claiming an already-assigned issue, not just a takeover. An abandoned run leaving a bare assignment would give the next run a lease carrying the old timestamp, so its fresh claim reads as expired immediately. Remove-then-add now covers both paths.
- **CLAIMED rule inside the fenced block** — correct, and my mistake: an imperative sat inside the digest's "exact shape" fence, so the surveyor could emit the instruction itself as a row. Moved to `Digest rules`.
- **Issue-less branch names** — correct gap: the template required `<issue>` while the contract still permits hotfixes and trivial fixes with no tracking issue. Those keep plain `claude/<area>-<desc>`; they go straight to a PR, so the PR body is the discoverable signal and no claim window applies.

## A note on where this stops

Four rounds, 18 findings, none of them noise — but the trend is worth naming honestly. Severity has narrowed (no P1 since round 1; this round was three P2s and a P3), while the *count* has not converged, and rounds 2–4 were each substantially **my own previous fixes creating new edge cases**. That is the documented behaviour of a prose/contract change rather than a code change: there is no compiler or test to bound the surface, so a sufficiently thorough reviewer can keep finding sharper phrasings indefinitely.

So my intent: **one more round to confirm this batch introduced nothing new, and then stop looping** — any remaining refinements get captured as a follow-up issue rather than another push. The substantive risks the review existed to catch were all found in rounds 1–2 (the out-of-scope-repo write, executing an external winner's code, the surveyor being blind to claim signals, the takeover never refreshing the lease); what is left is edge-case polish on a document, and it should not hold the change indefinitely.

The PR stays a draft either way — definition changes keep the human promotion gate, so @devantler decides when it lands.

@codex review

devantler:

All five fixed in `6ab02ae`. I said last round I'd stop if what came back was polish — it wasn't, so I didn't. Being explicit about that, because "I declared a stopping rule" is an easy way to wave off findings that actually matter.

**The sharpest one closes a hole I'd left in the core mechanism.** Two instances picking the same issue derive the **same deterministic branch name**, so my "re-verify immediately before the first push" settled nothing: both would see no branch and both would conclude they claimed it. That is exactly how #88 (52s) and #96 (135s) were lost, and the protocol written to prevent it still permitted it. The push now decides the race — a real commit on the claim branch so a losing push cannot be a trivial fast-forward, no force, then verify `git ls-remote` returns *your* sha. Rejected push or someone else's tip means you lost: stand down, never force over them.

The other four: takeover reusing a branch name the dead claim had already pushed commits to; the surveyor being blind to legacy claim branches (including any in flight during rollout); and `portfolio-maintenance` still teaching the no-op self-assign and still lacking the issue-less hotfix path. The last two are the same "fixed in the contract, not in the copies" class that has now bitten this PR three times — which is itself the argument for the definition-of-done rule about updating every agent file in the same change.

## Stopping rule, restated by severity rather than round count

Five rounds, 23 findings, none noise. My previous framing (one more round, then stop) was arbitrary and I'd have been wrong to apply it here. The defensible version:

- **P0/P1 — fix, however many rounds it takes.** These are the ones that would cause harm: rounds 1–2 produced a write to an out-of-scope repo, executing an external contributor's code, and a mechanism invisible to the survey that drives selection.
- **P2/P3 — one more pass, then capture the remainder as a follow-up issue** and let this land. Edge-case polish on a document should not hold a change indefinitely, and the marginal finding is now clearly narrower than the marginal delay.

By that rule this round earned its fixes (a residual race in the core mechanism is not polish), and the next round decides: anything P0/P1 gets fixed, anything P2/P3 goes to a follow-up issue with the specifics.

Still a draft — definition changes keep the human promotion gate, so @devantler decides when it lands.

@codex review

devantler:

All four fixed in `fdb3afe`. One was substantive and I'd have been wrong to wave it through on the P2 label:

**Claims are the agent account's, not "any assignee".** The surveyor reported an assignee *count*, so an issue assigned to a **human collaborator** (or Copilot) that also had a leftover `claude/*-<issue>` branch would satisfy the live-claim test — parking genuinely actionable work behind an unrelated person's assignment and timing the lease off the wrong assignment event. It now reports assignee **logins** and only a `devantler` assignment can be a claim. That also keeps the original *"respect assignees — do not do work others are assigned to"* direction intact, rather than quietly repurposing a human's assignment as an agent lease.

The other three were syncs of last round's fixes: the surveyor now matches takeover branches (`-<issue>-2`, …) which I introduced but left unmatchable; `portfolio-maintenance` carries the race-deciding push; the monorepo card carries the issue-less hotfix path.

## Applying the stopping rule

Round 6 was **all P2/P3 — no P0/P1** — which by the rule I set last round makes this the final pass. So:

- **If the next review returns P0/P1, I fix it** — that bar doesn't move.
- **If it returns P2/P3, I stop pushing** and file the remainder as a follow-up issue with the specifics, so this can land and the polish continues separately.

Six rounds, 27 findings, none noise. The pattern worth recording: rounds 2–6 were overwhelmingly *my own fixes creating new edge cases* in adjacent copies — three separate "fixed in the contract, not in the copies" recurrences. For a six-file amendment that is a structural property, not carelessness, and it is the strongest argument for landing this smaller next time: contract text in one PR, the surveyor/skill/card sync in another.

Still a draft; definition changes keep the human promotion gate.

@codex review

devantler:

**Stopping the review loop here.** Round 7 returned 6 findings, 0 of them P0/P1, which is the condition I set for stopping. I did not apply that mechanically, though — three of the six were fixed in `9ede906` and three are deferred, and the split is worth explaining because it deliberately overrides my own rule.

**Fixed (the label understated them):** two were the *contract-side* of the human-collaborator defect I fixed in the surveyor last round — the live-claim test treated **any** assignee as an agent lease, and the lease query took the newest `assigned` event for **anyone**. Shipping the surveyor fix without these would have left the authoritative file saying the wrong thing and the two disagreeing. An issue assigned to a human could still have been parked and then taken over after the window, which contradicts the standing *"do not do work others are assigned to"* direction. That is not polish regardless of the badge. The third was a one-line completion of my own previous round's takeover-branch fix, which I had applied to the surveyor's digest rules but not its branch scan.

**Deferred → devantler-tech/monorepo#2250:** takeover-tip verification naming the actual pushed branch, resolving the claim ref from the issue number rather than an assumed name, and repo-scoping the `#<issue>` body search. All three are real, none can strand a player, break a guardrail, or take someone's work.

## Closing summary

Seven rounds, 33 findings, none noise. Everything that could cause harm was found in rounds 1–2 and fixed: a write to an out-of-scope repo, executing an external contributor's code, a mechanism invisible to the survey that drives selection, a takeover that never refreshed its lease, the residual same-branch-name race that made the pre-push check useless, and treating a human's assignment as an agent lease.

The honest read on the volume: rounds 2–7 were overwhelmingly **my own fixes creating new edge cases in adjacent copies** — the contract, two skills, a card and an agent definition all restate these rules, so every correction has four echo sites and three separate "fixed in the contract, not in the copies" recurrences followed. That is a property of the change's shape, and the lesson is to land contract text and the copy-sync as separate PRs next time. It is recorded on #2250 as an acceptance criterion so the follow-up does not repeat it.

CI green, 0 unresolved threads, no further pushes planned. **Promotion is @devantler's call** — definition changes keep the human gate, and this one has now had considerably more scrutiny than most.

devantler:

Merged `main` in — it had moved 3 commits, including **#2239, which retires the definition-PR promotion gate**. That changes this PR's own path: it now self-promotes on the three genuine-readiness conditions like any product PR, rather than waiting on a separate gate.

One real conflict, in `portfolio-surveyor.md`: main added `--archived=false` to both org-wide searches while I added `assignees` to the issue search. Kept **both** — the searches now carry `--archived=false` AND the `assignees` claim signal, with main's explanatory note intact. Verified after the merge that my claim-protocol content (the section, skip reason (e), the exclusive-push rule, the devantler-only claim test) and main's new content both survived rather than assuming a clean auto-merge.

Re-requesting, since the merge moved the head and staled the previous green.

@codex review

devantler:

Three findings, 0 P0/P1. Two fixed, one deferred — and my own follow-up sweep found a fourth instance the review had not flagged.

**Honor human assignees when claims expire — fixed, and this is the one that mattered.** The run-loop copy still said to pick an issue up *"regardless of who's assigned"* once a claim lapsed, contradicting the contract change that only a `devantler` assignment is a claim. As written, an agent would take a **human collaborator's** assigned work after the 2h window — against the standing "do not do work others are assigned to" direction. Third recurrence of this class in this PR.

**Then I grepped every echo site instead of waiting for round nine** — and `product-engineering/SKILL.md` carried the same unqualified sentence, unflagged. Now fixed in all four (contract, both skills, surveyor), verified by sweeping for the phrase and confirming the human-collaborator carve-out is present in each.

**Digest row — fixed.** The shape said `assignees=<n>` (a count) while the step above requires logins, because only `devantler` is a claim signal. Now `assignee=devantler`.

**Claim-ref determinism — deferred**, already captured as item 2 of devantler-tech/monorepo#2250. The finding is right: the push only arbitrates the exact branch name a run derived, so two sessions choosing different `<area>-<desc>` stems still miss each other. It is a real residual, it is written down with the specifics, and it does not warrant another round here.

## Closing

Nine rounds, 36 findings, none noise. Every P0/P1 was fixed; what remains is one named residual on #2250. The dominant pattern, worth recording plainly: **this change has four echo sites, and the same defect recurred in them four times.** That is the argument for landing contract text and copy-sync separately, which is already an acceptance criterion on the follow-up.

The definition-PR gate was retired on `main` by #2239 while this was in review, so this now self-promotes on the ordinary readiness conditions rather than waiting. Promoting and merging once CI and a green review land at the new head.

devantler:

@codex review

PR #2239, opened 2026-07-18
docs(agents): retire the definition-PR human promotion gate

devantler:

## Why

Definition/self-improvement PRs were the last class still requiring your hand on the promotion button. Your direction (2026-07-18): injection is something we defend against **when inputs and prompts are read**, not at a gate downstream of a read that already went wrong. So the gate was costing friction on every definition change without being the place the defence actually lives.

## What

Definition PRs now follow the same path as product PRs — draft, hygiene pentad, the three genuine-readiness conditions, self-promote, merge. Updated the three constitution sites plus every restatement across the agent and skill files so nothing drifts apart.

The ingestion-side rules this now leans on (*Untrusted input*, NEVER-driven-by-repo-content) are marked load-bearing and tighten-only, since they carry the weight the gate used to share.

## One thing for you to decide

`AGENTS.md:1330` still reads: *"Loosening any guardrail requires the maintainer to direct and author it — you never propose it."* You directed this one but I authored it, so that clause is now contradicted by this very commit. I left it alone rather than widen scope on my own. Either it wants amending to match how you actually work, or it stays as-is and this PR is a one-off exception — your call, and I'll follow up whichever way you say.

Left as a draft deliberately: this PR grants the authority it would need to self-promote, and that authority isn't in effect until it merges. Not re-litigating — just noting the rule can't authorise its own adoption.

### DIFF
--- a/.claude/agents/daily-maintainer.md
+++ b/.claude/agents/daily-maintainer.md
@@ -18,8 +18,9 @@ lanes are unavailable, a qualifying agent self-review per contract *Autonomy →
 self-review*), **self-promote only
 on genuine readiness** (contract → *Autonomy*: programmatically tested + reviewed green + tried and
 evaluated as a user; maintainer direction 2026-07-16), then merge **directly** with bare
-`gh pr merge <n> --squash` — never `--auto`, which is bot-only. **The one exception: definition /
-self-improvement PRs keep the maintainer's promotion gate** (see the contract's *Self-improvement*).
+`gh pr merge <n> --squash` — never `--auto`, which is bot-only. **Definition / self-improvement PRs
+take this same path** — maintainer direction 2026-07-18 retired the separate promotion gate they used
+to keep (see the contract's *Self-improvement*).
 You drive *other* actionable trusted-author PRs to merge the same way — actionable single-author bots
 can arm `--auto`, but never via a branch-protection bypass; exact Renovate/Dependabot dependency PRs
 are automation-owned and receive no agent action. The maintainer steers after the fact via sessions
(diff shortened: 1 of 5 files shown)


devantler:

Amended in 972b97b, resolving the contradiction by **layer** rather than by dropping the requirement:

- **Prose/definition layer** — agent may author a loosening on your explicit direction, recording the direction and date in the text. This is what actually happened here, so the rule now matches practice.
- **Enforcement layer and that bullet itself** — stays your hand on the keystroke. Tightening it remains agent-direct and unchanged.

The `settings.json` edit for option 3 is prepared but not applied — it needs three coupled changes, not one, and details are in the session. Still a draft pending that.

devantler:

@codex review for correctness of the layered direct-vs-author rule and any place the definition still contradicts itself after retiring the definition-PR promotion gate.

devantler:

All four findings were valid and are fixed in d420255; threads resolved.

The channel-boundary one (AGENTS.md:1337) was the real catch — the "when he directs one" wording could have been satisfied by a `devantler` PR comment, since *Untrusted input* treats authenticated maintainer comments as instructions. That would have made repo content an authorization path for weakening a prose guardrail, precisely when the human promotion gate behind definition changes is being removed. Now explicitly interactive-session-only, with the comment carve-out stated as not extending to loosenings.

On AGENTS.md:1341 — correct that this PR is itself an agent-authored change to the rule it reserves. Rather than scope around it, the text now records that its own adoption was the exception it forbids, maintainer-directed in session on 2026-07-18, with the reservation binding every amendment after it.

@codex review — re-requesting at the new head d420255, since the previous green is staled by this push.

PR #2253, opened 2026-07-18
docs(agents): warn that the harness shell is zsh, which does not word-split

devantler:
> 🤖 Generated by the Daily AI Engineer (interactive session, at @devantler's request)

## Why

The most common tool error in the agents' own telemetry is `argument required when using the --repo flag` — 41 hits in three days. It looks like a flaky `gh` call, so it gets retried rather than fixed.

It is not flaky. The Bash tool's shell is **zsh**, which does not word-split unquoted expansions, so the bash idiom for looping over "repo number" pairs quietly loses the PR number. `gh` then blames the flag that *is* present, which is why the real fault has stayed invisible.

This wastes a call and a retry every time, and it lands hardest in the per-run trusted-PR sweep — first-priority work, and exactly where these loops get written.

## What

Documents the trap in the latency-discipline section, with three working alternatives, and generalises it to every implicit-splitting idiom.

Measured rather than guessed: pairing each failing result with the command that produced it across 250 sessions shows **24 of 24** such failures used this idiom and **zero** used literal arguments. RED/GREEN proved in zsh 5.9, then the fixed loop was run against real `gh` across three repositories.

This is the first fix produced by the agent-improver routine (#2251) — found by mining the agents' own telemetry rather than by noticing it by hand.

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -1115,6 +1115,18 @@ window, unnoticed. The work was never the bottleneck; the **scheduling** was.
   re-verify **once** — not a fix-one/re-run round trip per finding.
 - **Parallelize independent setup.** Clones, subagents, and independent investigations start
   together in the background, not one after another.
+- **The Bash tool's shell is `zsh`, which does NOT word-split unquoted expansions — never sweep a
+  PR list with `set -- $pair`.** The bash idiom for iterating `"repo number"` pairs silently
+  collapses: `for pr in "ksail 6045" …; do set -- $pr; gh pr view $2 --repo devantler-tech/$1` leaves
+  `$1` holding the *whole* string and `$2` **empty**, so `gh` runs with no PR number and fails
+  `argument required when using the --repo flag`. The flag is present — the positional argument in
+  front of it vanished, which is why the error misdirects. Measured: **24 of 24** such failures
+  across 250 sessions (2026-07-14→18) used this idiom; **zero** used literal arguments. It is not a
+  flaky call, it is a shell mismatch, and it hits hardest in the per-run trusted-PR sweep where these
+  loops are written most. Use one of:
+  `IFS=' ' read -r repo n <<< "$pr"` (portable), `set -- ${=pr}` (zsh's explicit split flag), or
+  simply **write the calls out** — two plain `gh` lines beat a clever loop and stay readable.
+  The same trap applies to every implicit-splitting idiom (`for x in $(cmd)`, `cmd $args`).
 
 This changes only *ordering and overlap* — never the quality bar. Validation, RED/GREEN proof,
 root-cause fixing, and every guardrail are unaffected; the point is to stop paying for them serially.
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -1115,8 +1115,8 @@ window, unnoticed. The work was never the bottleneck; the **scheduling** was.
   re-verify **once** — not a fix-one/re-run round trip per finding.
 - **Parallelize independent setup.** Clones, subagents, and independent investigations start
   together in the background, not one after another.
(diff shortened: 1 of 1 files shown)


devantler:
> 🤖 Generated by the Daily AI Engineer (interactive session)

@codex review

devantler:
> 🤖 Generated by the Daily AI Engineer (interactive session)

## Both findings valid — fixed in 1ee9813

I verified all three of your claims empirically before acting, and every one held:

```
bash:  set -- $pair    → $1=ksail  $2=6045          (works — nothing to fix)
bash:  set -- ${=pair} → bad substitution           (hard syntax error)
zsh:   set -- $pair    → $1='ksail 6045'  $2=''     (the actual bug)
zsh:   for x in $(cmd) → [aa] [bb] [cc]             (splits — different hazard)
```

**P2 — scope to zsh.** This was the serious one, and you're right about why: `AGENTS.md` is the canonical **cross-tool** file, and the Codex sibling runs bash. Written globally, my rule would have told a bash-backed agent that working code was broken and handed it `${=pair}` — a hard syntax error there. I'd introduced a harness-specific fact into a shell-agnostic contract, which is worse than the bug I was fixing, since it would have broken the *other* instance.

Now scoped to zsh, leading with the `IFS read` form (verified correct in both shells) as the default, with `${=var}` explicitly marked zsh-only and called out as unsafe under a bash shebang.

**P3 — separate command substitution.** Also correct, and a genuine semantics error on my part: zsh *does* IFS-split `$(cmd)`; only parameter expansion is exempt. Grouping them would have sent agents chasing the wrong root cause when a `$(...)` loop was merely hitting ordinary whitespace splitting. Split out, with the parameter-expansion family named precisely (`set -- $var`, `cmd $args`).

Re-requesting review at the new head.

@codex review

devantler:
> 🤖 Generated by the Daily AI Engineer (interactive session)

**Valid — fixed in 4a8e7e8.** Verified before acting:

```
dash: IFS=" " read -r a b <<< "x y"   → Syntax error: redirection unexpected
sh/bash/zsh: repo=${pr%% *}; n=${pr##* }  → repo=[ksail] n=[6045]  (identical in all three)
```

You're right, and the irony is precise: the paragraph tells agents to use this form **when the shell is unknown**, so an advertised portable fallback that dies under `/bin/sh` would have recreated the exact cross-shell breakage this rule exists to prevent — one shell over from the original bug.

Now leads with pure parameter expansion, which is POSIX and verified identical in `sh`, `bash` and `zsh`. The `read` form stays documented as a bash/zsh equivalent with its limit stated explicitly, so nobody reaches for it under a `#!/bin/sh` shebang.

That is three rounds where my *generalisation* was the defect rather than the finding — instance→class, then zsh→all-shells, now bash/zsh→POSIX. Worth recording as a pattern rather than three separate fixes.

Re-requesting review at the new head.

@codex review

devantler:
> 🤖 Generated by the Daily AI Engineer (interactive session)

**Valid — fixed in 379fd31.** You caught that my previous fix was half-applied: I replaced the code block with the POSIX form but left a trailing sentence still sending agents to the `read` form when the shell is unknown — precisely the case where its here-string breaks. The paragraph contradicted itself, and anyone reading to the end landed on the wrong advice.

Now points to parameter expansion, the only one of the three with no shell restriction.

That is the same generalisation-scope defect a third time in this file alone, so it is now encoded as a standing check in the improver skill rather than fixed once more by hand.

@codex review

devantler:
> 🤖 Generated by the Daily AI Engineer (interactive session)

All review findings addressed in `434ca5f`; threads resolved. Re-requesting review at the new head.

@codex review

PR #2254, opened 2026-07-18
docs(agents): stop the surveyor diagnosing review-lane outages

devantler:

## Why

A survey this run reported a "fresh both-lane outage" across 12 PRs and flagged the agent self-review fallback as likely warranted. It was wrong — Codex review objects existed on at least three of those PRs, one at the current head. The actual condition was ordinary: greens staled by pushes.

Acting on that digest would have meant self-reviewing a dozen PRs a real reviewer had already covered. That is a quality-gate bypass reached by accident, from a conclusion the surveyor is not equipped to draw.

## What

The surveyor now reports per-PR state only and is explicitly barred from diagnosing a lane outage or suggesting the self-review precondition — that inference belongs to the orchestrator, which has the per-lane evidence. Also names the three ordinary causes of a `none` row so it is read rather than escalated.

The per-PR vocabulary needed no change; it was already correct.


devantler:

@codex review for whether this rule can be satisfied vacuously, and whether barring the surveyor from lane diagnosis leaves any legitimate signal unreportable.

devantler:

Both surveyor findings valid; fixed in 0fcac59, threads resolved.

Also corrected a mistake of mine: this PR was branched off `claude/ingestion-egress-hardening` rather than `main` while targeting `main`, so it carried all of #2240's commits. That is why two findings landed here against `AGENTS.md` content that belongs to #2240 — they are fixed there (b079c06). Retargeted to stack on #2240, so this PR is now the one-concern surveyor change it was meant to be.

- **`none` was satisfiable by assertion** — the row carried no observable proof both surfaces were checked, so the orchestrator could not tell a real absence from the filter miss the rule exists to prevent. It now reports the artifact counts actually seen plus the abbreviated head matched against, and a non-zero count beside `none` is defined as a contradiction to investigate rather than a row to emit.
- Banning outage *diagnosis* also removed the channel for legitimate per-lane *evidence*, which the fallback contract needs. Added a neutral factual `lane_signal=` row: report what the reviewer said, never characterise it.

@codex review — re-requesting at 0fcac59.

devantler:

All three valid; fixed in e48fff7, threads resolved.

All three were one defect: I introduced vocabulary without wiring it into the "exact shape" the digest is declared to follow — so a surveyor holding real evidence would have had to drop it or go off-contract.

- `lane_signal` had no row in the fenced digest. Added one with repo/PR context.
- The evidence suffix was described in prose while the row definitions still read `green_review=...|none`, and the same digest legitimately uses `rd=none` for GitHub's `reviewDecision`. Scoped explicitly to `green_review`, both row definitions updated.
- The contradiction rule counted any bot comment — but a walkthrough summary, command reply, or rate-limit notice is not review output, and `green_review=none` beside those is correct. Scoped the counts to actual review output so those PRs still get a valid row rather than being hidden.

@codex review — re-requesting at e48fff7.

devantler:

Both valid; fixed in aee63c0, threads resolved.

- The run-loop skill that **consumes** this digest still documented the bare `none` token, so the surveyor and orchestrator contracts disagreed on the row this PR introduces. Updated it with the evidence-bearing form, the `green_review`-only scoping, and the stale-is-normal note — the agent-file sync rule applies to a producer/consumer pair like this one.
- The contradiction rule counted any review-output artifact, but a findings review at a previous head with no current result is the *common* review-needed state — the rule would have suppressed that row entirely, hiding exactly the PRs that need a re-request. Narrowed to current-head artifacts; stale counts beside `none` are the re-request signal.

@codex review — re-requesting at aee63c0.

devantler:

Both valid; fixed in ab632fa, threads resolved, consumer skill synced.

- The `LANE-SIGNAL` template told the surveyor to quote a reviewer's notice **verbatim** into the digest — relaying untrusted text straight into the orchestrator's context, which violates this repo's own untrusted-input and delimiter-safe-quoting rules. I wrote a rule that broke a rule I wrote two PRs ago. Now requires summarising in the surveyor's own words with mention and command tokens neutralised.
- `rev`/`cmt` aggregated both bots, so one lane's stale artifact could mask the other lane's missed one — precisely the Codex findings-review miss this evidence exists to catch. Split per lane, consumer contract updated to match.

@codex review — re-requesting at ab632fa.

devantler:

Valid; fixed in 5fb719a, thread resolved.

I split the evidence per lane in the prose and in the consumer skill but left all four **row definitions** on the aggregate form — so the surveyor's own template contradicted both its explanation and the orchestrator contract, and following the template would have emitted exactly the ambiguous signal this change exists to remove. A partial find-and-replace, which is the same shape as the P1 on #2240: I applied a change to the instance in front of me rather than to every occurrence.

All occurrences are now per-lane.

@codex review — re-requesting at 5fb719a.

devantler:

Valid; fixed in 75e40cb2c4, thread resolved.

Fourth instance of a single root cause this session: I wrote a bare `green_review=none` in the lane-signal prose while the rule and all four row templates require the evidence-bearing form — so a surveyor copying that compact token drops the suffix and reintroduces the ambiguous absence signal.

Rather than fix the one occurrence, I added the general rule: prose saying "none" names the *state*; the emitted *token* always carries the per-lane suffix. That closes the class.

@codex review — re-requesting at 75e40cb2c4.

devantler:

Rebased onto `main`, which had moved — siblings landed #2234 and #2204 while this was in review. Verified the branch now carries **only** the surveyor change plus its consumer sync, with `memory-hygiene.sh`, its test, and the ci.yaml changes all intact.

@codex review — re-requesting at 438c1cffce, since the merge staled the previous green.

PR #2277, opened 2026-07-19
feat(ai-engineer): add the Gemini instance as the third scheduled engineer

devantler:
## Why

You asked for Gemini in the scheduled rotation for more delivery speed. The contract still described **two** instances and an hourly sweep, so all three agents would have booted reading a wrong picture of their own deployment — including a cadence line and a collision-risk section that are actively load-bearing.

## What

Records the third instance and carries the change through everywhere two was assumed: cadence (now a ~40-minute sweep, while per-instance carry-forward stays ~2 hours), the claim protocol's collision evidence, the two-writer branch warning, host least-privilege scope, and both stewardship rules. Adds a `GEMINI.md` shim so a default Gemini CLI finds the canonical instructions.

**Worth your attention:** Gemini CLI has no scheduler of its own, so that instance runs off a machine-local `launchd` agent rather than a native scheduled task — a deliberate asymmetry the agent-improver is now told not to flatten. It also runs `--yolo` per your call, and upstream reports its policy engine being ignored headless, so for that instance the constitution is the guardrail rather than the runtime.

The machine-local half (runtime config, loader, wrapper, schedule) is already in place and is not version-controlled. It is **live but idle** until the Google sign-in is done — one interactive `gemini` login.


devantler:
Closing unbuilt: the maintainer decided not to add the Gemini instance yet (2026-07-19). The machine-local runtime, loader, wrapper and launchd schedule have all been removed, so the constitution correctly continues to describe two instances.

Nothing here is lost if we revisit — GitHub keeps the branch restorable from this closed PR.

Issue #2108, opened 2026-07-09
ci(coderabbit): docstring-coverage pre-merge check false-positives block draft promotion portfolio-wide


devantler:
> 🤖 Generated by the Daily AI Assistant

## Problem

CodeRabbit's `Docstring Coverage` pre-merge check (org-level, currently `mode: error`) under-counts docstrings: it only credits a docstring that appears **inside the PR diff hunk**. So any PR that (a) changes the *body* of an already-documented function, or (b) uses an anonymous closure (which can't carry a docstring), scores far below the 80% threshold even when the code is fully documented.

Concrete evidence: ksail#5978's `pkg/client/kubeconform/client.go` is 100% documented (every func/type has a doc comment) yet scored **33.33%** and blocked promotion. The maintainer won't promote drafts with red pre-merge checks — so this silently wedges draft promotion across **every** repo that hits it.

## Proposed direction

Restore CodeRabbit's own default (non-blocking `warning`) for docstring coverage — Go doc-comment coverage is already enforced by golangci-lint in CI, so nothing is lost. Two levers:
- **Per-repo** `.coderabbit.yaml` (`reviews.pre_merge_checks.docstrings.mode: warning`) — proven in ksail#5994; propagate to platform + other active repos.
- **Or org-level** — flip the setting in the CodeRabbit org dashboard so no per-repo file is needed (not version-controlled; maintainer action).

## Rough size

S — one small config file per active repo (or one org-dashboard toggle). Blocked on validating ksail#5994 lands and clears #5978's check first.

Reference: ksail#5994 (ksail fix), ksail#5978 (evidence).

devantler:
> 🤖 Generated by the Daily AI Assistant

## Status + recommendation

**ksail interim fix landed:** #5994 merged the per-repo `.coderabbit.yaml` (`docstrings.mode: warning`) to ksail `main`.

**Key operational finding (corrects an earlier assumption):** merging the per-repo config to `main` does **not** retroactively clear the docstring pre-merge check on already-open PRs. CodeRabbit reads `.coderabbit.yaml` from the PR's head/merge state, so each open draft that predates the config still shows the stale ❌ until its branch is **updated to include the file** (confirmed on ksail#5978: still ❌ after #5994 merged; cleared only after an `update-branch`). This means the per-repo approach carries an ongoing tax — every open draft on every repo must be branch-updated to benefit.

[...]

PR #2285, opened 2026-07-19
ci(coderabbit): keep docstring coverage advisory instead of merge-blocking

devantler:

## Why

CodeRabbit's Docstring Coverage pre-merge check scores one coverage percentage across everything a
PR touches, and it cannot be limited to particular files. This repository's executable code is the
agent tooling under `.claude/scripts` — 17 bash scripts holding around 95 shell functions — which do
not carry doc comments, because shell scripts are not written that way.

The result is that a PR changing those scripts is marked as failing a documentation standard it was
never in breach of, and a failed pre-merge check blocks the merge. We ship changes to those scripts
constantly, so this sits directly in the path of routine work.

## What

Adds a CodeRabbit config that keeps Docstring Coverage as advisory rather than merge-blocking. The
coverage number is still reported; it just stops holding up merges.

The same change was already made for `ksail` and `.github` for the same reason. This is the third
repository, not a new policy.

Part of #2108

### DIFF
--- a/.coderabbit.yaml
+++ b/.coderabbit.yaml
@@ -1,0 +1,30 @@
+# yaml-language-server: $schema=https://coderabbit.ai/integrations/schema.v2.json
+#
+# CodeRabbit configuration. Only the settings that differ from CodeRabbit's defaults are
+# listed here; everything else stays at the default.
+#
+# inheritance: without this, a repository .coderabbit.yaml is the sole config source and
+# REPLACES any organization/UI settings (the docstring ERROR gate itself lives in the UI,
+# since the YAML default is `warning`). Enabling inheritance keeps every UI/org setting —
+# path filters, request_changes_workflow, custom checks — and only overrides the one value
+# defined below, so this change stays scoped to docstrings.
+inheritance: true
+reviews:
+  pre_merge_checks:
+    # Docstring Coverage measures a single coverage percentage across a PR's changed files
+    # and cannot be scoped per-path (the only path mechanism, reviews.path_filters, would
+    # drop the files from the whole review rather than from just this metric).
+    #
+    # This repository's executable code is the agent tooling under .claude/scripts — 17 bash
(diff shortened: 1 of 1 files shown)


devantler:

@codex review

Adds a `.coderabbit.yaml` keeping Docstring Coverage advisory rather than merge-blocking, matching
`ksail` and `.github`.

Worth focusing on: `inheritance: true` is the load-bearing line. Without it a repository
`.coderabbit.yaml` **replaces** the org/UI configuration instead of merging with it, so the file
would silently drop every other org setting while looking like a one-value change. Schema-checked
against `schema.v2.json` (`inheritance` boolean; `docstrings.mode` enum `off|warning|error`, YAML
default `warning`).

devantler:

Readiness record.

**Programmatically tested.** All required checks green at `56004c7`. The config was validated against
CodeRabbit's published `schema.v2.json`, not assumed: `inheritance` is a top-level boolean whose
documented meaning is *"use parent settings for any values not defined here"* with default **false**,
and `docstrings.mode` is an enum of `off|warning|error` whose YAML default is `warning`. That default
is what confirms the blocking `error` gate is a UI-level setting, which is the whole reason this file
is needed.

**Reviewed.** Codex green at `56004c736a` — the current head — with no findings.

**Tried and evaluated as a user, with one honest limitation.** I deliberately did *not* request a
CodeRabbit review here as proof: this PR's diff contains no functions, so the Docstring Coverage
metric would report *"No functions found… Skipping"* and pass vacuously. That would have looked like
confirmation while testing nothing — the exact failure mode this PR exists to fix.

The behavioural evidence is that the identical file is already live in `devantler-tech/.github`,
where CodeRabbit itself reported the check as `⚠️ Warning` and stated it was *"non-blocking under the
refreshed configuration"* (`.github#103`) — the same mechanism, same two settings, observed working.

One operational note carried over from that repo: an incremental `@coderabbitai review` on an
unchanged head does **not** re-read configuration. A `full review` is required after a
`.coderabbit.yaml` change.

Issue #2308, opened 2026-07-20
fix(surveyor): a Codex clean-pass at head must outrank an older findings review object


devantler:
> 🤖 Generated by the Daily AI Engineer

## Evidence

This tick the digest reported `green_review=codex-findings@<old-sha>` (→ **NEEDS-FIX**) for four own drafts that were in fact **green at head**. Verified live against `issues/<n>/comments`:

| PR | head | Codex clean-pass comment | digest said |
|---|---|---|---|
| `ksail`#6267 | `248223bc6e` | `248223bc6e` ✅ | `codex-findings@dd4b5f8998` |
| `ksail`#6279 | `e137ae58de` | `e137ae58de` ✅ | `codex-findings@8fc2598ad4` |
| `agent-plugins`#72 | `5e5ee3afc8` | `5e5ee3afc8` ✅ | `codex-findings@e50a2c08e9` |
| `platform`#2635 | `72917dde95` | `72917dde95` ✅ | `codex-findings@14f2c2a751` |

In every case a **findings review object at an older sha** coexisted with a **newer clean-pass issue comment naming the current head**, and the older object won.

## Affected audience and impact

Every run that consumes the digest. A PR reported `codex-findings` is classified **NEEDS-FIX**, so the run either re-requests a review it already has or treats the draft as unfinished. Review quota is **per-account and contended by ~7 parallel sessions**, and a wasted CodeRabbit request costs ~1h of lane availability — so this converts an already-green PR into repeated spend. It also inflates apparent NEEDS-FIX load, hiding genuinely blocked work.

## Root cause

[`.claude/agents/portfolio-surveyor.md`](.claude/agents/portfolio-surveyor.md) (~L172) directs a sweep of `issues/<n>/comments` **and** `pulls/<n>/reviews` for "the latest actual Codex review output". The two surfaces carry different timestamp fields (`created_at` vs `submitted_at`) and Codex uses different shapes per outcome — findings are a review **object**, a clean pass is an issue **comment** with `**Reviewed commit:**` and no `commit_id`. "Latest" is therefore undefined across them, and the implementation resolves it toward the review object.

## Expected behaviour

Head-match is decisive and evaluated **before** any recency comparison: a clean-pass naming `headRefOid` is `codex@<head>` even when a findings review object exists at an older sha. Only findings **at head** yield `codex-findings`.

## Acceptance criteria

[...]

PR #2322, opened 2026-07-20
feat(agents): add Cursor Bugbot as a review lane, priority Codex > Cursor > CodeRabbit

devantler:

## Why

We had two review lanes, and when Codex ran out of credits that left exactly one — CodeRabbit, throttled to roughly one review an hour and contended by every parallel session. The result is visible in today's survey: **100 open own/trusted PRs, none with a green review at head.** Review capacity, not engineering throughput, is what's gating the portfolio.

## What

Adds **Cursor Bugbot** as a third review lane and sets the priority to **Codex → Cursor → CodeRabbit**, per your direction today. Triggered by a `bugbot run` PR comment.

One thing worth knowing, because it shapes the change: Bugbot reports a clean review as a **status check**, not as a review object or a comment like the other two. Anything looking only at reviews and comments simply cannot see it, and would report "no review" on a PR Bugbot had already passed — the same blind spot we just fixed for Codex in #2308/#2309. So the surveyor learns that surface here, rather than after it starts misreporting.

Also kept deliberately narrow: this grants Bugbot **reviewer** standing only. It does **not** make `app/cursor` a trusted PR author — that decision is still open in #2297. Both roles share the `cursor[bot]` login, so the gate keys on the check-run rather than the login; otherwise our own Cursor Automation instance could appear to approve its own PRs.

The self-review fallback now requires **all three** lanes down, so it becomes correspondingly rarer.


devantler:
> Requested by the 🤖 Daily AI Engineer

cursor review

devantler:
> Requested by the 🤖 Daily AI Engineer

bugbot run

devantler:
> Requested by the 🤖 Daily AI Engineer

@cursor review

devantler:
@cursor review

devantler:

All 4 Bugbot findings addressed in `696bdab` — every one was valid, and all four shared a single root cause: I updated the headline rules to three lanes but left downstream sites saying two.

- **High — surveyor self-review gate still two-lane:** fixed. It now requires a per-lane failure line for all three lanes; a `self@` token can no longer be emitted with Bugbot never tried.
- **AGENTS.md self-review template:** heading and pre-merge note re-framed to three lanes.
- **SKILL `(e)` omitted Bugbot check-runs:** the classification step now names the check-run surface explicitly, matching the `bugbot@` token it already emitted.
- **SKILL still said "both lanes"/"both reviewers":** fixed.

Why my own sweep missed them: I grepped `both lanes` and `CodeRabbit or Codex`, but not `CodeRabbit **and** Codex`, `both-lanes-down`, or `both reviewers`. Re-swept exhaustively across all variants — now clean. All 9 contract tests pass.

Also corrected the trigger and artifact shape in this PR to what was *measured* rather than what Cursor's docs claim — the documented `bugbot run` / `cursor review` commands did not fire, and neither did `@cursor review` with a disclosure line above it.

Re-requesting review at the new head.

(The next comment is a bare `@cursor review` — Bugbot exact-matches the whole comment body, so the disclosure has to sit in its own comment. That carve-out is documented in this PR.)

devantler:
@cursor review

PR #2345, opened 2026-07-21
docs(ai-engineer): read a file at the worktree path before editing it

devantler:

## Why

Every run creates a throwaway worktree, and the agents then routinely try to edit a file there without reading it first — because they *did* read it, in the main checkout, minutes earlier. The tool refuses the edit, the run pays a wasted round-trip, and it happens often enough to be the **largest single tool-error signature in both the 1-day and 7-day telemetry windows**. It has sat at the top of four consecutive scorecards without anyone naming the cause.

## What

Adds a short rule to the *Execution model* section: a fresh worktree is a fresh path, so the first touch of each file there is a `Read`. The evidence is in the text so a future run can re-verify rather than trust it.

Measured 2026-07-14→21 on the Claude instance: 134 refusals (73 under this monorepo), 126 on a path never read in that session, **102 of those — 81% — inside a `.claude/worktrees/` path**. Scoped explicitly to the Claude lane, since the sibling runtimes record no error flag and this could not be measured for them.

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -1625,11 +1625,12 @@ slug only for `gh` commands). **Submodule worktree isolation breaks whenever a s
 initialised** — a stray shared `core.worktree` makes `git worktree add` resolve back into the main
 checkout, silently collapsing every parallel session into one physical tree.
 
-**A fresh worktree is a fresh PATH — `Read` a file THERE before your first edit of it.** Read-tracking
-is keyed on the **absolute path**, so every file in a newly-created worktree counts as unread even when
-you read the same file in the main checkout minutes earlier — the edit is refused with *"File has not
-been read yet"* and the run pays a wasted round-trip. Knowing a file's contents is not the same as
-having read it *at the path you are editing*; the path, not the filename, is the identity. Measured
+**A fresh worktree is a fresh COPY — `Read` a file THERE before your first edit of it.** Reading a file
+in the main checkout does **not** count as having read the worktree's copy: it is a different file on
+disk and the read record does not carry across, so the edit is refused with *"File has not been read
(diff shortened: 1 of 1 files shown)


devantler:

Requesting **Cursor Bugbot** (lane 2 — Codex is credit-exhausted account-wide, re-probed 02:1xZ). CI is green at head `03239af`.

Pre-submission self-review is done and it changed the diff: the first draft asserted that read-tracking is "keyed on the absolute path". That mechanism was **inferred, not tested**. The hardlink probe I ran to settle it was **vacuous** — its own control (an unread, distinct-inode copy) was edited successfully, so the guard never fired in either arm and neither result bears on the question. Second commit withdraws the mechanism claim and leaves only the observed behaviour, which is what the 102 measurements actually support.

devantler:
@cursor review

devantler:

**Readiness record** (head `03239af`):

1. **Programmatically tested** — CI green, every non-reviewer check `SUCCESS`/`SKIPPED`; 0 unresolved threads; 0 Bugbot inline findings.
2. **Reviewed** — **Cursor Bugbot green**, check-run `Cursor Bugbot` `conclusion: success` at this head. CodeRabbit was not asked (lane 3, and lane 2 answered), so `premerge=not-posted` is the lane-choice consequence, not a gap.
3. **User-evaluation** — this change is constitution prose consumed by agents, so it has **no exercisable runtime surface**; per the contract I am saying so explicitly rather than implying a behavioural check. What *was* exercised is the evidence underneath it: the mining run that produced the 134/126/102 split, re-verified against full paths after the first pass (truncated to 60 chars) returned the inverted answer.

Honest limit worth recording: the fix asserts nothing about *why* the read record does not carry across, only that it does not. My attempt to establish the mechanism produced a vacuous result and was withdrawn rather than reported as a finding.

Issue #2344, opened 2026-07-21
CodeRabbit's rate-limit commit status blocks every own PR from ever reaching CLEAN


devantler:
> 🤖 Generated by the Daily AI Engineer

## Evidence

When CodeRabbit is rate-limited it posts a **commit status** on the PR head:

    context: CodeRabbit   state: failure   description: "Review rate limit exceeded"

A failing commit status — required or not — puts the PR in `mergeStateStatus: UNSTABLE`. It can therefore never read `CLEAN`, which is the state the engineering contract requires before a trusted-author PR is merged.

Measured on 2026-07-21 04:1xZ, four of the ten own open drafts were blocked by nothing else:

| PR | pentad | blocker |
|---|---|---|
| `monorepo#2290` | CI green, 0 threads, pre-merge green | `CodeRabbit=failure` → UNSTABLE |
| `monorepo#2287` | CI green, 0 threads | same |
| `actions#661` | CI green, 0 threads, green review at head | same |
| `gitops-tenant-template#94` | CI green, 0 threads, pre-merge green, green review at head | same |

It is not new: the same block was recorded on 2026-07-20 (world-at-ruin#320 held back for it) and again the following tick. The status is a **service quota signal, not a review verdict** — it says nothing about the change.

## Who this affects

Every agent lane, on every repo. It converts a third-party throttle into an indefinite merge block on work that is otherwise finished, and the wait is unbounded because the quota is shared across all our repos and all three agent instances.

## Expected behaviour

A CodeRabbit throttle should not change a PR's mergeability. Either CodeRabbit stops publishing quota state as a failing commit status, or we stop letting it reach the merge gate.

## Candidate levers (to be evaluated, not pre-decided)

1. A `.coderabbit.yaml` setting that suppresses the review commit status.
2. An org/repo-level setting for the CodeRabbit GitHub App's status reporting.
3. If neither exists, treat it as an upstream defect and report it — a quota message is not a check result.

## Acceptance criteria

- A rate-limited CodeRabbit no longer leaves an otherwise-finished PR outside `CLEAN`.
- Demonstrated on a real PR: throttle present, `mergeStateStatus` still `CLEAN`.
- Whatever the chosen lever, it is applied consistently across the portfolio, not one repo.

Rough size: S–M (config change plus verification), or a small upstream report.

devantler:
> 🤖 Generated by the Daily AI Engineer

[...]

PR #2351, opened 2026-07-21
docs(agents): let a local review round stand in when a lane is rate or billing limited

devantler:

## Why

Your direction today: *"We likely need to allow local review rounds when external review providers are rate or billing limited, such that we are not blocked by it."*

Until now a review provider saying "try again in 34 minutes" counted as a working reviewer, so finished work had to sit and wait for it. This morning all three providers were unavailable at once — one out of credits, one out of Cursor spend, one throttled — and nothing could merge at all, on changes that had no outstanding concerns.

## What

A review the agent runs itself now counts when no provider will deliver, including when the reason is simply that we have hit a rate or spend limit. The same principle applies to the merge gate: CodeRabbit posts a red mark on the commit when it is throttled, which was holding otherwise-finished pull requests out of merge for over a day — that mark reports its own service state, not a judgement about the change, so it no longer blocks. Deliberately narrow: nothing else red is ever waved through.

The standard itself does not move — same bar as a provider's review, posted the same way, and it only counts if it is clean on the exact commit being merged. Six checks lock that down.

Part of #2344 — this stops the throttle blocking us, but the root-cause fix there is still the CodeRabbit setting that makes it post a red mark at all, so that issue stays open.

### DIFF
--- a/.claude/agents/daily-maintainer.md
+++ b/.claude/agents/daily-maintainer.md
@@ -13,9 +13,9 @@ product, working from the one monorepo checkout where each product is present as
 responsible for keeping every product healthy *and* moving it forward. You act directly with the `gh`
 CLI and `git`, and — as a **trusted author** — you **drive your own PRs to merge yourself**: work in
 a draft, drive the full current-head hygiene pentad clear (green required checks and pre-merge checks,
-no review findings/conflict, and a green Codex, Cursor Bugbot or CodeRabbit review at that head — or, when ALL THREE
-lanes are unavailable, a qualifying agent self-review per contract *Autonomy → Fallback — agent
-self-review*), **self-promote only
+no review findings/conflict, and a green Codex, Cursor Bugbot or CodeRabbit review at that head — or, when no
+lane will deliver at that head (unavailable, or rate/billing limited), a qualifying local review
+round per contract *Autonomy → Local review round*), **self-promote only
 on genuine readiness** (contract → *Autonomy*: programmatically tested + reviewed green + tried and
 evaluated as a user; maintainer direction 2026-07-16), then merge **directly** with bare
 `gh pr merge <n> --squash` — never `--auto`, which is bot-only. **Definition / self-improvement PRs
--- a/.claude/agents/portfolio-surveyor.md
+++ b/.claude/agents/portfolio-surveyor.md
@@ -419,7 +419,7 @@ paginate) and say so, rather than surveying a partial list.
 **Report per-PR state; never diagnose a portfolio-level condition from it.** You are a reporter, and
 several of the states you emit look alarming in aggregate without being so. Specifically: **never
 conclude that a review lane is down, stalled, rate-limited, or outaged, and never suggest the
-*Fallback — agent self-review* precondition is met** — that inference is the orchestrator's alone,
+*Local review round* precondition is met** — that inference is the orchestrator's alone,
 it requires per-lane evidence you do not gather, and acting on it wrongly means self-reviewing PRs a
 reviewer already covered.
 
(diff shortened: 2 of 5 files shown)


devantler (commented):

## Local review round — clean

**Reviewed commit:** `8e68767356b467052e442a04d65f5ac3400c9850`

### Lane evidence (why this review is local)

| Lane | State at review time |
|---|---|
| Codex | out of credits account-wide since 2026-07-20; no window |
| Cursor Bugbot | `usage limit reached` — Cursor spend exhausted, no retry window; an admin must raise it |
| CodeRabbit | rate-limited (`Review limit reached`, 34-minute window stated at 06:52Z). It also structurally never `APPROVE`s a finding-free diff, so even a served review would not produce a gate-satisfying artifact here |

This PR is itself the change that makes the above a valid basis for a local round. Reviewing it under the rule it introduces is circular, so I am being explicit about the authority: the direction, not the unmerged text, is what licenses it.

### What I checked, and what it turned up

- **Does it weaken anything beyond what was directed?** No. The trigger moves; the standard does not. Same bar as a bot lane, still posted as a real Review with resolvable threads, still only counts clean at a sha equal to the current head — and three of the six new guards exist specifically to pin that down (`the bar, only the trigger`, `clean at a sha`, and the pre-submission disambiguation, which is unchanged and still passing).
- **Is the merge carve-out bounded?** This was the part worth being suspicious of, because "ignore a red signal" is exactly how a gate rots. It is scoped to one context — the review provider's own quota status — and the sentence `never a red CI check, never a required check, never a finding` is guarded so it cannot be quietly generalised. Note the consequence, which I then hit within the hour: `platform#2769` has a red MegaLinter check and is therefore **not** mergeable under this rule. The rule constrained me on its first real test, which is the behaviour I wanted from it.
- **Does the stale-`mergeStateStatus` exception open a hole?** It requires *every* check-run and status on the head to be `success`/`skipped`, so a still-pending required check (conclusion `null`) does not qualify, and the merge API remains the authority — it enforces every real rule and refuses cleanly. Measured on `actions#661`: 120/120 green, `UNSTABLE`, and `PUT /pulls/661/merge` succeeded first try.
- **Propagation.** The old section name and the "all three lanes unavailable" phrasing appeared at seven sites across four files. All seven are updated; `grep` for the old strings returns nothing. This is my recurring partial-propagation defect, so I counted rather than assumed.
- **Guards.** Six new assertions, each RED-proven against `origin/main`. One of them initially failed for the right reason: I had pinned a phrase that the prose line-wraps, so it could never match — re-anchored to a single-line phrase and re-proven.

### Two things this review changed

1. **`Fixes #2344` → `Part of #2344`.** The PR body over-claimed. #2344's acceptance criteria are about the status not being posted at all, which is a CodeRabbit configuration fix (`reviews.fail_commit_status`); this PR only stops that status blocking us. Closing the issue would have dropped the root-cause work.
2. A stuttering phrase left by the rename in the contract test's header comment (`the substitute-for-a-bot-lane SUBSTITUTE for a bot review`), fixed in `8e68767`.

devantler:

**Readiness record at `8e68767`.**

**1. Programmatically tested.** All checks green, `mergeStateStatus: CLEAN`, 0 unresolved threads. Both contract suites pass — `self-review-contract.test.sh` and `portfolio-surveyor.test.sh` — and the six new assertions are each RED-proven against `origin/main`.

**2. Reviewed.** Local review round posted above, clean at this head, with the per-lane evidence table. It produced two real changes rather than a rubber stamp: the `Fixes #2344` over-claim corrected to `Part of`, and a stuttering phrase left by the rename.

**3. Tried and evaluated as a user.** I used the new rule on three PRs before promoting this one, which is the only way to find out whether it actually resolves the situation it was written for:
- **monorepo#2348** — merged `5469636d` after two local rounds (round 1 found a P1 and it was fixed).
- **gitops-tenant-template#94** — merged `ec7966ca`. It had been parked for two ticks on nothing but CodeRabbit's frozen rate-limit status; that is precisely the case this change addresses.
- **platform#2769** — **not** merged. It has a red MegaLinter check, which this rule explicitly does not excuse. The carve-out constrained me on its first real test, which is the outcome I wanted from it, and the underlying red baseline is filed as platform#2780.

Promoting and merging.

PR #2458, opened 2026-07-25
fix(scripts): compare worktree paths by identity, not spelling

### LINKED ISSUE #2457: fix(scripts): submodule-init --check falsely reports ISOLATION BROKEN on a case-only path difference

## Evidence

`submodule-init.sh --check` reports `libraries/agent-plugins` as **ISOLATION BROKEN** on a tree where it
is in fact correctly isolated, and exits non-zero:

```
libraries/agent-plugins — ISOLATION BROKEN: existing linked worktree
'…/monorepo/.Codex/worktrees/agent-plugins-claude-desktop-83' resolves to
'…/monorepo/.codex/worktrees/agent-plugins-claude-desktop-83', not its own path.
```

The two paths differ **only in the case of one character** (`.Codex` vs `.codex`). On the agent host's
case-insensitive APFS volume they are the same directory — measured:

```
inode .codex: 102677667
inode .Codex: 102677667
worktree inode via .codex=127602832  via .Codex=127602832   ==> SAME DIRECTORY
```

Root cause is the comparison at `.claude/scripts/submodule-init.sh:87-90`: both sides are normalised
with `pwd -P`, which resolves **symlinks but not case**. `want` keeps the spelling recorded in the
worktree admin `gitdir` file (`.Codex/…`, written by a Codex-lane session), while `got` comes back from
`git rev-parse --show-toplevel` in git's own spelling (`.codex/…`), so the string compare fails on a
difference that does not exist on disk.

Every agent run on this host, on the run's **pre-flight** path. Two harms, and the second is the worse
one:

1. **It fails closed in the wrong direction.** The check tells the run "do not edit it" and exits
   non-zero for a submodule that is perfectly safe to edit, so legitimate mandated work on
   `agent-plugins` is blocked for a phantom reason.
2. **It cries wolf against the real thing.** A genuine isolation break silently collapses parallel
   sessions into one working tree — the single most damaging failure this script exists to catch. A
   warning that fires routinely on a false positive trains runs to discount the true positive.

The trigger is not exotic: any sibling lane whose worktree root is recorded in a different case
reproduces it, and `.codex/` is a standing directory on this host.

## Expected behaviour

Two paths that resolve to the same directory on the host filesystem compare equal, so a case-only
difference is not reported as broken. A **genuine** collision — a stray `core.worktree` pointing at a
different tree — must still be reported exactly as it is today.

## Acceptance criteria

- [ ] `submodule-init.sh --check` reports `libraries/agent-plugins` as isolated on the current tree and
      exits 0 (with no other submodule regressing).
- [ ] Paths are compared by **filesystem identity** rather than by string, so a case-only difference on
      a case-insensitive volume compares equal, and the comparison stays correct on a case-sensitive
      volume too.
- [ ] A RED/GREEN guard in `.claude/scripts/*.test.sh` proves the false positive is gone **and** that a
      genuinely colliding worktree is still reported — the negative control matters more than the
      positive one here.

**Size:** small (a comparison fix plus its guard).

devantler:

## Why

The pre-flight isolation check every agent run depends on was raising a false alarm: it declared one
submodule unsafe to edit when it was perfectly fine. That blocked legitimate work for a phantom
reason — and worse, a warning that fires routinely teaches the reader to ignore the one that flags a
genuine collision, which is the failure this check exists to catch.

## What

The check now decides whether two paths are the same directory by asking the filesystem instead of
comparing the text, so two spellings of one directory no longer read as a conflict. A real collision
is still reported exactly as before, and the guard covering that case is unchanged.

Fixes #2457

### DIFF
--- a/.claude/scripts/submodule-init.sh
+++ b/.claude/scripts/submodule-init.sh
@@ -28,13 +28,6 @@ die() {
 
 warn() { printf 'submodule-init: %s\n' "$1" >&2; }
 
-# NEVER compute a submodule's gitdir — ask git. Run from a linked superproject worktree (the documented
-# execution model for agent runs), git puts each submodule's gitdir under
-# `.git/worktrees/<super-wt>/modules/<path>`, NOT under `.git/modules/<path>` — and that is where the
-# stray `core.worktree` lands too. Any assumed path is wrong exactly where this script matters most.
-super_root=$(git rev-parse --show-toplevel) || die 'not inside a git repository'
-cd "$super_root"
-
 # The gitdir git is ACTUALLY using for this submodule.
 module_dir() { git -C "$1" rev-parse --path-format=absolute --git-common-dir 2>/dev/null; }
 
@@ -44,12 +37,27 @@ module_dir() { git -C "$1" rev-parse --path-format=absolute --git-common-dir 2>/
 module_tree() { (cd "$1" 2>/dev/null && pwd -P); }
 
 # Do two paths name the SAME directory? Compare filesystem IDENTITY (device+inode, via `-ef`), never
-# the spelling. `pwd -P` resolves symlinks but NOT case, so on a case-insensitive volume — APFS, the
-# default on the macOS agent host — one directory reached as `.Codex/…` and reported back by git as
-# `.codex/…` string-compares unequal while being the same inode. That produced a false ISOLATION
-# BROKEN on a correctly-isolated submodule, which is worse than merely noisy: it blocks edits to a
-# safe tree AND trains the reader to discount the warning that flags a real collision.
-# Fails closed — an empty or non-existent path is never "the same directory".
+# the spelling.
+#
+# Two independent reasons the spellings legitimately differ, and BOTH are load-bearing:
+#
+# 1. CASE, and this is shell-specific — measured, because the wrong shell hides it. Under **bash**
+#    (this script's shebang, 3.2.57 on the host) `pwd -P` resolves symlinks but does NOT fold case:
+#    `cd alpha/beta && pwd -P` prints the typed `alpha/beta`. Under **zsh** the same command prints
+#    the on-disk `Alpha/Beta`. So on a case-insensitive volume (APFS, the host default) a worktree
+#    recorded as `.Codex/…` stays `.Codex/…` here while git reports `.codex/…` — one inode, two
+#    strings, and the old `[ "$got" != "$want" ]` fired. Traced live on `libraries/agent-plugins`:
+#      got  = …/.codex/worktrees/agent-plugins-claude-desktop-83   (git, lowercase)
+#      want = …/.Codex/worktrees/agent-plugins-claude-desktop-83   (bash pwd -P, case preserved)
+#    Reproducing this in zsh shows both sides folded and NO mismatch — verify in bash, not the
+#    interactive shell.
+# 2. An unreadable worktree makes BOTH sides empty (`-d` passes, `cd` and `git -C` both fail), and
+#    `[ "" = "" ]` is TRUE — so without the emptiness guard below this returns "same directory" and
+#    `--check` reports an unverifiable worktree as isolated. That fail-open predates this helper.
+#
+# A false positive here is a safety regression, not just noise: it blocks edits to a safe tree AND
+# trains the reader to discount the warning that flags a real collision. So fail CLOSED — an empty or
+# non-existent path is never "the same directory".
 same_dir() {
   [ -n "${1:-}" ] && [ -n "${2:-}" ] || return 1
   [ "$1" = "$2" ] && return 0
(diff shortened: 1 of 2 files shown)


devantler:

## User evaluation

Exercised as the consumer of this guard — a run doing its pre-flight isolation check — at head `29e44316`.

**The reported symptom, before and after, on the real 21-submodule tree.** Original script from
`origin/main`:

```
libraries/agent-plugins — ISOLATION BROKEN: existing linked worktree
  ".../.Codex/worktrees/agent-plugins-claude-desktop-83" resolves to
  ".../.codex/worktrees/..." (×2 worktrees)
```

With this branch: `--check` exits **0**, **21 isolated ✓, 0 broken**. The submodule that was declared
unsafe to edit is now correctly reported safe, and no other submodule changed verdict.

**The mechanism, traced rather than assumed.** `bash -x` on the original showed the actual operands:

```
got  = .../.codex/worktrees/agent-plugins-claude-desktop-83   (git)
want = .../.Codex/worktrees/agent-plugins-claude-desktop-83   (bash pwd -P — case preserved)
```

🔑 This is **shell-specific, and checking it in the wrong shell hides it entirely**: under `bash` (the
shebang) `cd alpha/beta && pwd -P` prints the typed `alpha/beta`; under `zsh` it prints the on-disk
`Alpha/Beta`. Reproducing in an interactive zsh shows both sides folded and no mismatch — which is
exactly how this reads as a no-op if measured there. Verified `bash` 3.2.57, `zsh`, and `/bin/sh`
separately.

**RED/GREEN by ablation.** Reverting `same_dir` to the old `[ "$1" = "$2" ]` fails precisely the two
identity assertions while the three fail-closed/negative controls still pass — the correct signal,
since those controls exist to stop the fix degrading into a blanket "equal" and so should be
insensitive to it. Separately confirmed the emptiness guard is load-bearing: without it
`same_dir "" ""` returns **true**, i.e. two unresolvable paths compare as the same directory.

**Cross-platform, and the skip is honest.** CI ran the suite on both runners: ubuntu passes the
symlink-alias assertion and prints an explicit `SKIP` for the case-only one (case-sensitive FS);
macOS passes the case-only assertion. So the platform where the defect is reachable covers it, and
the other platform still covers the comparison.

**Sourcing is side-effect-free**, which the first commit got wrong: sourced from `/tmp` (a non-git
directory) the shell survives, `cwd` is unchanged, and `same_dir` is defined.

Pre-submission self-review: an adversarial subagent pass produced three real corrections, all folded
in (`29e4431`) — the shell-specific case claim, the untested both-empty fail-open, and the guard
placement. Its headline claim that the fix was a production no-op was **refuted by measurement**: it
had measured `pwd -P` in zsh. That review also surfaced a separate fail-open now filed as #2460.

Not asserted end-to-end on purpose: `probe` prunes an unverifiable worktree before the sweep that
should flag it (#2460), so an end-to-end assertion would pin that bug rather than this fix.

devantler (commented):

**Reviewed commit:** `29e4431` (head at review time)

### Per-lane state that justified reviewing locally

| Lane | State | Evidence |
|---|---|---|
| CodeRabbit | rate-limited | `Review limit reached` ×3 today at 06:28:59Z / 12:08:53Z / 12:33:48Z (last was `full review` to escape the incremental wedge); windows escalated 14 → 20 → **57 min**, so the account quota is spent for the period, not briefly throttled |
| Codex | usage-limited | 12:38:47Z — `reached your Codex usage limits for code reviews`, **no retry window**; needs account credits |
| Cursor Bugbot | spend-limited | 12:42:15→17Z — check-run `neutral` + `output.title=Error`, **0** inline findings, failed in **2s**; `cursor[bot]` reports a Cursor usage/spend limit, admin-only to lift |

All three are provider-quota states rather than judgements about this diff. Per the 2026-07-21
direction, a rate limit *with* a stated window counts as not-delivering, so the local round applies.
Neither Codex nor Bugbot was re-requested at this head: both state no retry window, so re-requesting
is waste.

### What I actually verified (not a restatement of the diff)

- **The reported symptom, end to end.** Original script from `origin/main` on the live 21-submodule
  tree: `ISOLATION BROKEN` ×2 on `libraries/agent-plugins`. This branch: **21 isolated ✓, 0 broken,
  exit 0**. No other submodule changed verdict.
- **The mechanism, traced under the script's own interpreter.** `bash -x` showed
  `got=.../.codex/...` (git) vs `want=.../.Codex/...` (bash `pwd -P`). Independently confirmed the
  shell split: `bash` and `/bin/sh` preserve the typed case, `zsh` folds it. This is the load-bearing
  fact — measured in zsh, the whole fix reads as a no-op.
- **Ablation.** Reverting `same_dir` to `[ "$1" = "$2" ]` fails exactly the two identity assertions;
  the three fail-closed/negative controls still pass, which is correct — they bound the fix rather
  than test it.
- **The emptiness guard is load-bearing and now covered.** Without it `same_dir "" ""` returns
  **true**, so two unresolvable paths compare as one directory. That fail-open predates this helper.
- **Invocation matrix after moving `super_root`/`cd` below the sourced guard** — the one structural
  risk in the diff: `--check` from the repo root, from a linked superproject worktree, and from a
  subdirectory all behave as before; no-args still exits 1 with usage; a non-git directory still dies.
  The linked-worktree run reports 0 submodules — **controlled against the original script from the
  same directory, which reports 0/0 identically**, so that is inherent to an unpopulated worktree and
  not a regression.
- **Sourcing is side-effect-free**: sourced from `/tmp`, the caller survives, `cwd` is unchanged, and
  the helper is defined. The first commit got this wrong; `29e4431` fixes it.
- CI green on **both** runners (28 checks, 0 failures); `shellcheck` clean on both files.

### Scope discipline

An end-to-end assertion for the emptiness guard is deliberately **absent**: `probe` runs
`git worktree prune` before `check_existing_worktrees`, so an unverifiable worktree's admin entry is
deleted before the sweep meant to flag it. That is a separate fail-open, filed as #2460. Asserting the
end-to-end behaviour here would pin that bug rather than this fix, so the guard is covered at helper
level with a pointer.

**Verdict: no P0/P1 findings**

PR #2449, opened 2026-07-25
fix(surveyor): exclude Renovate Dependency Dashboard from oldest-actionable

devantler:
> 🤖 Generated by the Agentic Engineer

## Why

Renovate's Dependency Dashboard (`platform#313`) is an open issue by design and was heading every platform oldest-actionable ranking forever, so each run re-derived that it is not real work.

## What

Extends the automation-owned carve-out to issues (exact `renovate[bot]` / `dependabot[bot]` authors only), teaches the surveyor to exclude them from Advance ranking, and guards the contract so it cannot silently regress.

Fixes #2349

PR #2463, opened 2026-07-25
fix(scripts): verify existing worktrees before pruning, and scope the prune

### LINKED ISSUE #2460: bug(scripts): submodule-init --check prunes unverifiable worktrees, then reports isolated (fail-open)

## Evidence

`submodule-init.sh --check` reports a submodule **isolated ✓** when one of its linked worktrees cannot
be verified at all — the fail-open class the script's own comments say it must never have.

Reproduced on a throwaway fixture (a `file://` submodule with one linked worktree, made unsearchable
with `chmod 0400`, so `-d` passes while both `cd` and `git -C` fail):

```
admin dirs BEFORE check: sub-wt
-d sub-wt?      yes
cd works?       no
git -C works?   <fails>
--- run --check ---
submodule-init: sub — isolated ✓
admin dirs AFTER check: ls: super/.git/modules/sub/worktrees/: No such file or directory
```

**Root cause — ordering inside `probe()`.** It runs, in this order: `resolves_to_itself` → add a
throwaway probe worktree → compare → `worktree remove --force` → **`git worktree prune`** → `rm -rf` →
**then** `check_existing_worktrees`. `git worktree prune` deletes admin entries whose working tree is
missing *or* cannot be verified, so the unverifiable worktree's entry is gone by the time the sweep
meant to flag it runs. The sweep then finds nothing to check and the submodule is declared isolated.

Two distinct problems in that one call:

1. **Fail-open.** The evidence is destroyed by the checker itself, so "no linked worktrees to verify"
   and "a linked worktree that cannot be verified" become indistinguishable — and the second is
   reported as success.
2. **`worktree prune` is not scoped to this script's own probe entry.** The header states `--check`
   "prunes its own admin entry", but `prune` is repository-wide. Any *other* session's worktree that is
   momentarily unreadable — restrictive permissions, an unmounted volume, an in-flight `chmod` — has its
   admin entry deleted, which permanently breaks that worktree. For a tool whose entire purpose is
   protecting parallel sessions from each other, `--check` damaging a live sibling's worktree is the
   wrong direction.

Found while fixing #2457 (path comparison). The comparison fix is correct and separate: with the
emptiness guard, `same_dir "" ""` returns false as intended — the empty pair simply never reaches it,
because the entry is pruned first. #2457's PR therefore covers the guard at helper level and points
here rather than asserting the end-to-end behaviour, which would pin this bug.

Every agent run: this is the pre-flight trust gate that decides whether a submodule is safe to edit. A
false **isolated ✓** is materially worse than the false positive #2457 fixes — a colliding or
unverifiable worktree gets edited, and cross-session collisions silently corrupt parallel sessions'
work, which is the failure this script exists to prevent.

## Expected behaviour

- An admin entry that cannot be verified is **reported**, never silently pruned: unverifiable must fail
  closed and be distinguishable from "no linked worktrees".
- `check_existing_worktrees` observes the admin state **before** anything prunes it.
- Any prune stays scoped to this script's **own** probe entry (remove that entry directly rather than
  a repository-wide `prune`), so a sibling session's momentarily-unreadable worktree is never deleted.

## Acceptance criteria

- [ ] With an unsearchable linked worktree present, `--check` exits non-zero and names it; it never
      reports the submodule isolated.
- [ ] A sibling worktree's admin entry still exists after `--check` runs, even when that worktree was
      unreadable during the run.
- [ ] Pruning is scoped to the script's own `probe-iso-*` entry.
- [ ] A guard in `.claude/scripts/submodule-init.test.sh` covers both, including the negative control
      that a genuinely stale entry is still cleaned up (so the fix does not just disable pruning).

**Size:** small–medium (reorder + scope the prune, plus guards).

devantler:

## Why

`submodule-init.sh --check` is the pre-flight gate every agent run uses to decide whether a submodule is safe to edit. It could report a submodule **isolated ✓** when one of its linked worktrees could not be verified at all — the exact fail-open the gate exists to prevent. A colliding or unverifiable parallel session would then be edited, silently corrupting another session's work.

It also ran a repository-wide prune, which could delete a *sibling* session's bookkeeping whenever that session's files were briefly unreadable — permanently breaking the very parallel session the tool is meant to protect.

## What

The check now inspects existing worktrees **before** it cleans anything up, so an unverifiable one is reported and the check fails closed. Cleanup is scoped to the check's own throwaway worktree instead of sweeping the whole repository.

Fixes #2460

### DIFF
--- a/.claude/scripts/submodule-init.sh
+++ b/.claude/scripts/submodule-init.sh
@@ -91,8 +91,13 @@ resolves_to_itself() {
 # leave one `$gitdir/worktrees/*` entry pinned at the shared checkout, and that session keeps colliding
 # while a fresh probe stays clean. `--check` is the trust gate for exactly this parallel-session case,
 # so it must enumerate the live worktrees too. Read-only: it only rev-parses other sessions' trees.
+# $2 (optional) is THIS probe's own admin dir, excluded by filesystem IDENTITY. It must be an exact
+# identity match, never a name pattern: an earlier version skipped any worktree whose PATH contained
+# the substring `/probe-iso-`, so a genuinely unverifiable sibling living under a directory merely
+# NAMED that way was silently skipped and the submodule reported isolated ✓ — the same fail-open class
+# as #2460, one level down. A sibling's admin entry is never excluded.
 check_existing_worktrees() {
-  local path=$1 rc=0 mdir wtadmin wt got want
+  local path=$1 skip_admin=${2:-} rc=0 mdir wtadmin wt got want
   # Iterate the LINKED-worktree admin dirs under the submodule gitdir directly, rather than
   # `git worktree list`: the submodule's MAIN checkout legitimately resolves to itself, and so does a
   # colliding linked worktree whose stray `core.worktree` points at the shared checkout — so
@@ -102,9 +107,12 @@ check_existing_worktrees() {
   [ -d "$mdir/worktrees" ] || return 0
   for wtadmin in "$mdir"/worktrees/*/; do
     [ -f "$wtadmin/gitdir" ] || continue
+    # Skip ONLY this probe's own admin dir, by identity — see the header note above.
+    if [ -n "$skip_admin" ] && [ -d "$skip_admin" ] && [ "${wtadmin%/}" -ef "$skip_admin" ]; then
+      continue
+    fi
     # The `gitdir` admin file points at the worktree's own `.git` file; its parent is the worktree.
     wt=$(dirname "$(cat "$wtadmin/gitdir" 2>/dev/null)")
-    case "$wt" in *"/probe-iso-"*) continue ;; esac
     # A pruned/missing worktree dir cannot host a live colliding session — skip it.
     [ -d "$wt" ] || continue
     got=$(git -C "$wt" rev-parse --show-toplevel 2>/dev/null) || got=''
@@ -175,6 +183,15 @@ probe() {
     return 1
   fi
 
+  # Capture the probe's REAL admin dir up front. `git worktree add` does NOT guarantee the admin dir
+  # is named after the worktree: when `worktrees/<name>` already exists it counter-appends (adding
+  # `probe-iso-1-2` against an existing entry yields `probe-iso-1-21`, measured on git 2.55.0). Then
+  # `worktree remove` matches by PATH and drops the counter-appended one, so removing "$name" by name
+  # afterwards would delete a DIFFERENT — possibly live — worktree's entry, destroying exactly the
+  # sibling this scoping exists to protect. Resolve it, and remove that.
+  local probe_admin=''
+  probe_admin=$(git -C "$probe_dir" rev-parse --absolute-git-dir 2>/dev/null) || probe_admin=''
+
   # Resolve both sides to physical paths (/tmp vs /private/tmp, symlinked checkouts) before comparing.
   local got want rc=0
   got=$(git -C "$probe_dir" rev-parse --show-toplevel 2>/dev/null) || got=''
@@ -195,18 +212,21 @@ probe() {
   # evidence this sweep reads: "no linked worktrees to verify" and "a linked worktree that cannot be
   # verified" became indistinguishable, and the second was reported as isolated ✓ (#2460). The sweep
   # skips this probe's own `probe-iso-*` entry, so running it first is safe.
-  check_existing_worktrees "$path" || rc=1
+  check_existing_worktrees "$path" "$probe_admin" || rc=1
 
   # Clean up ONLY this probe's own entry. A repository-wide `git worktree prune` also deletes a
   # SIBLING session's admin entry whenever that session's tree is momentarily unreadable (restrictive
   # permissions, an unmounted volume, an in-flight chmod) — permanently breaking the parallel session
   # this script exists to protect. `worktree remove` already drops the admin entry on success; the
-  # scoped rm is the fallback for when it does not.
+  # scoped rm is the fallback for when it does not, and it targets the RESOLVED admin dir rather than
+  # a name that git may have counter-appended.
   git -C "$path" worktree remove --force "$name" >/dev/null 2>&1 || true
   rm -rf "$probe_dir"
-  local mdir
-  mdir=$(module_dir "$path") || mdir=''
-  [ -n "$mdir" ] && rm -rf "${mdir:?}/worktrees/${name:?}"
+  if [ -n "$probe_admin" ]; then
+    case "$probe_admin" in
+      */worktrees/*) rm -rf "${probe_admin:?}" ;;
+    esac
+  fi
 
   [ "$rc" -eq 0 ] && printf 'submodule-init: %s — isolated ✓\n' "$path"
   return "$rc"
--- a/.claude/scripts/submodule-init.sh
+++ b/.claude/scripts/submodule-init.sh
@@ -17,8 +17,10 @@
 #
 # `--check` never modifies submodule content, tracked files, or other sessions' worktrees, but it is
 # NOT strictly read-only: to prove isolation empirically it adds and then removes a throwaway,
-# uniquely-named probe worktree (and prunes its own admin entry). That empirical add/remove is the
-# whole point — it catches a dangling `core.worktree` a config read alone would miss.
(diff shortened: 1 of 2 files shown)


devantler:

## Validation

**RED/GREEN on the reported defect.** Fixture: a `file://` submodule with one linked worktree made unsearchable (`chmod 0400`), so `-d` passes — the entry is not skipped as "missing" — while both `cd` and `git -C` fail, i.e. genuinely unverifiable.

| | `origin/main` | this branch |
|---|---|---|
| exit code | **0** | **1** |
| verdict | `sub — isolated ✓` | `ISOLATION BROKEN … resolves to '<unresolvable>'` |
| sibling admin entry after the run | **destroyed** | **survives** |

**Self-test: 29 assertions, 0 failures.** RED-proved against `origin/main`: exactly **4** failures, all four of the new #2460 guards, with the 22 pre-existing assertions still passing — so the harness is sane and the guards fail for the reason under test, not because the script would not run. The three `cleanup:` assertions **pass in both directions by design**: they are negative controls proving the fix does not simply disable cleanup.

One correction found while proving this: my first version of the cleanup control used `find` on a path that does not exist on a fully-cleaned tree, which trips `pipefail` and aborted the suite before the last two assertions ran. Replaced with a `nullglob` array count; all 29 assertions now demonstrably execute.

## User evaluation

Exercised as the tool's actual consumer — the pre-flight gate every agent run uses — against the **real 21-submodule portfolio** in the populated main checkout:

- `--check` → **21 isolated ✓ / 0 broken**, exit 0, 3.9s.
- **Sibling sessions' worktree bookkeeping preserved**: `platform` 20 → 20 entries, `applications/ksail` 3 → 3. This is the property the repository-wide prune put at risk; those entries include live parallel sessions.
- **0** `probe-iso-*` leftovers, so scoping the cleanup did not leak state.

I deliberately did **not** run the pre-fix script against the shared checkout to demonstrate the destructive half: it would prune live sibling sessions' admin entries for real. That half is proven on the throwaway fixture instead.

devantler:

## Adversarial review — 2 real defects found and fixed

I ran an adversarial pass tasked with **refuting** this change, requiring every finding to be reproduced rather than reasoned. It found two genuine fail-opens, both in code this PR *promotes into a load-bearing position*. Fixed in `265fe7d`.

**1 — the probe self-exclusion matched a path SUBSTRING.** The sweep skipped any worktree whose path contained `/probe-iso-`. While the sweep ran *after* cleanup that filter was effectively dead code; moving the sweep first made it the only self-exclusion, so a genuinely unverifiable sibling living under a directory merely *named* `probe-iso-experiments` was silently skipped and the submodule reported `isolated ✓` — the same fail-open class as #2460, one level down. Now matched by **filesystem identity** against this probe's own admin dir.

**2 — `git worktree add` counter-appends, so the admin dir is not reliably `$name`.** With `worktrees/probe-iso-1-2` already present, adding `probe-iso-1-2` creates `probe-iso-1-21` (measured, git 2.55.0); `worktree remove` then matches by path and drops the counter-appended one, so removing `$name` by name deletes a *different* — possibly live — worktree's entry. The reviewer demonstrated a live sibling being destroyed while `--check` still printed `isolated ✓`. Now the probe's real admin dir is resolved up front and that is what gets removed.

Both are exactly the harm this PR's own comments claim to prevent, reintroduced by a different mechanism. That is why the pass runs even when I expect to refute it.

**Evidence now: 37 assertions, 0 failures, 0 skips.**

| baseline | failures | which |
|---|---|---|
| `origin/main` | **4** | the original #2460 guards |
| my previous commit `56ed9bd3` | **2** | the two new self-exclusion guards; all 4 originals still pass |

**Honest limit on coverage.** The counter-append hazard needs an exact `probe-iso-$$-$RANDOM` collision, which cannot be forced deterministically from the test, so it is pinned by invariant rather than RED-proof: with `git worktree remove` forced to fail via a PATH shim, the scoped removal alone cleans up **and** a pre-existing sibling is asserted still present *and still functional*. The reviewer's own mutation testing is what showed the earlier version of that control could not distinguish the two cleanup mechanisms.

**Also fixed from the same review:** the cleanup assertions were pure *absence* checks that would pass if the helper never ran at all, so they now carry a positive precondition that the probe actually executed.

**Refuted and not changed:** `${mdir:?}`/`${name:?}` genuinely abort on unset; no state leaks from dropping the repo-wide prune (12 runs, both layouts, including with `worktree remove` forced to fail); the `chmod 0400` fixture is non-vacuous and does not SKIP on either CI runner; no regression to the 22 pre-existing assertions.

**User evaluation re-run at this head** against the real 21-submodule portfolio: **21 isolated ✓ / 0 broken**, sibling bookkeeping unchanged (platform 20→20, ksail 3→3), 0 probe leftovers. `shellcheck -S style` clean.

devantler:


Reserving the CodeRabbit lane at head `265fe7d28c`. CodeRabbit is demonstrably serving again — it delivered a real review object on #2449 at 14:37:04Z after its announced window elapsed. Both metered lanes recorded hard usage limits earlier today (Codex weekly, Bugbot monthly Cursor spend), so they are not requested.

devantler:


@coderabbitai full review

devantler (commented):

**Reviewed commit:** `265fe7d28c1bdcf35147f1e3837e772717c4e5d1`

**Per-lane evidence, measured — not inherited:**

| Lane | State at/for this head | Measured |
|---|---|---|
| **CodeRabbit** | requested at this exact head; responded with `rate limited by coderabbit.ai` — no review object | 14:39:39Z today |
| **Codex** | `usage limit` for code reviews, **no retry window stated** (weekly quota) | 12:38:47Z today |
| **Cursor Bugbot** | check-run `neutral` + `output.title: Error`, `usage limit reached` — the run never started; monthly Cursor **spend** cap, admin must raise it | 12:42:17Z today |

CodeRabbit is genuinely serving today — it delivered a real review object on #2449 at 14:37:04Z — so this is a true one-review-per-window throttle rather than a dead lane. I deliberately did not re-spend Codex or Bugbot: a usage/spend cap states no window and cannot succeed on retry, so firing them would only post comments that fail.

### Verdict: no P0/P1 findings

This is not a rubber stamp. An adversarial pass over this same diff earlier **found two real P1 fail-opens**, both now fixed in `265fe7d` and recorded in a PR comment. What follows is the independent re-derivation I did on top of that.

**Checked and satisfied:**

1. **Single call site updated.** `check_existing_worktrees` gained a second parameter; there is exactly one caller (`probe`, line 215) and it passes `$probe_admin`. No stale-arity call remains.
2. **The degraded path fails safe.** If `rev-parse --absolute-git-dir` fails, `probe_admin` is empty and the sweep runs with no exclusion. I exercised that directly: the probe's own entry resolves to itself, so `same_dir` passes and it is **not** falsely flagged (`rc=0`). An empty skip therefore costs nothing and never produces a false ISOLATION BROKEN — which matters because a false positive here is itself a safety regression.
3. **The scoped removal cannot widen.** `[ -n "$probe_admin" ]` short-circuits, `${probe_admin:?}` aborts on unset, and the `*/worktrees/*` case guard means a `probe_admin` that resolved to a *main* gitdir (no `/worktrees/` segment) is skipped rather than removed. Every failure mode declines to delete.
4. **Identity, not spelling.** The new self-exclusion uses `-ef`, consistent with `same_dir`'s existing identity rule, so a symlinked or differently-spelled path cannot smuggle a sibling past the sweep.
5. **Ordering is safe.** Running the sweep before cleanup cannot flag the probe itself (see 2), and the probe entry is the only thing that ordering could have raced.

**Honest limits of the evidence, stated rather than glossed:**

- The counter-append hazard (P1 #2 in the adversarial pass) needs an exact `probe-iso-$$-$RANDOM` collision, which the suite cannot force deterministically. It is pinned by invariant — with `git worktree remove` forced to fail via a PATH shim, the scoped removal alone cleans up **and** a pre-existing sibling is asserted still present *and still functional* — not by RED proof.
- The `chmod 0400` fixture is skipped for anyone running as root. It is verified non-vacuous on both CI runners (all four #2460 guards **run**; they do not SKIP), so CI coverage is real.

**Programmatic state:** 37 assertions / 0 failures / 0 skips. RED-proved: 4 failures against `origin/main`, 2 against the prior commit `56ed9bd3`, with all pre-existing assertions still passing in both. `shellcheck -S style` clean. CI green (12 success, 16 skipped, 0 failures).

Issue #2302, opened 2026-07-20
Cross-lane claim races are never arbitrated — a lane-neutral claim ref, with four proven traps


devantler:
> 🤖 Generated by the Daily AI Engineer

Split out of #2292 on a pre-committed trigger: the claim mechanism drew a new P1 in each of two consecutive review rounds and started requiring changes in a third file, so it ships on its own rather than growing that PR further.

## Evidence — the hole is PRE-EXISTING, not introduced by the third lane

The claim protocol's rule 4 settles a race by having both instances derive the *same* branch name, so one non-force push is refused. That premise is already false: each instance writes its own namespace, and `codex/*` has been in live use alongside `claude/*` for some time — **measured 2026-07-20: 17 `codex/*` PRs on ksail, 3 on world-at-ruin, 3 `cursor/*` on monorepo**. Two *different* instances racing one issue therefore both push cleanly and both believe they won. Cross-lane races have never been arbitrated; adding a third lane widens an existing hole rather than creating one.

The recorded cost of unarbitrated races is high: six end-to-end builds discarded in ~24 hours on world-at-ruin, one lost by 52 seconds and one by 135.

## Design

Arbitrate on a lane-neutral `agent-claim/<issue>` ref that every instance derives identically from the issue number alone, **before** creating its lane-specific work branch.

## Four traps, all proven by execution — do not re-derive these

1. **Arbitration works** — two clones, second writer's non-force push is rejected, `git ls-remote` returns the winner's sha. ✅
2. **Never judge the push by its exit status.** `git push … | tail` returns **exit 0 on a rejected push** (the pipeline reports `tail`'s status). Only the `ls-remote` tip comparison is safe.
3. **Identical claim commits collide.** Every instance commits as `devantler`, so a fixed message on a shared base in the same second yields a **byte-identical** commit — reproduced exactly: two clones both produced `e0c73d68fed2075216b62023f8147c411929d3da`, identical down to the signature. Both pushes succeed, both read the tip as theirs, arbitration silently fails in the close race it exists for. A nonce fixes it. ⚠️ A first fixture did *not* reproduce this because the clones had different parents — anchor both on `origin/main` or the negative result looks like refutation.
[...]

PR #2465, opened 2026-07-25
feat(agents): order work by a single ladder — open PRs, drafts included, before new work

devantler:

## Why

We keep starting work faster than we finish it. On 2026-07-25 there were **99 open PRs the agent
authored — every single one still a draft, none ever promoted**, median age **6.9 days**. Eighteen of
them were already mergeable and had been sitting that way for about five days; sixteen had drifted
into conflict; forty-nine of the eighty-eight sampled were never touched again after the hour they
were opened.

Throughput was not the problem — roughly 27 of these merged per day that same week. The problem is
that finished work was not being picked back up, so the pile grew underneath it.

The cause turned out to be a wording gap rather than a missing rule. The contract's top-priority item
was "drive PRs to merge", but elsewhere that set was defined as *non-draft* PRs — so the agent's own
drafts matched no priority at all, and the "finish before you start" rule sat far away in another
section, phrased as advice about opening new drafts. Checked against live state, the old top priority
matched **one** PR across the whole portfolio and **none** of the agent's own.

## What

Replaces three drifting copies of the per-run order with **one work-selection ladder**, worked
top-down: **breakage → open PRs (including our own drafts) → security issues → bugs → oldest issue**.

- Our own drafts are now explicitly the top rung, and the `non-draft` wording is scoped to the merge
  step rather than the sweep — closing the hole that produced the pile.
- Severity now outranks age in the issue queue. **18 open security issues and 99 bugs** were
  previously competing on age alone against 368 open issues.
- A new CI check keeps the ladder from drifting again: nine negative controls, each verified to
  actually mutate the text before its result is trusted.

One judgement call worth flagging: live breakage stays *above* open PRs, as it already was. A red
`main` blocks every PR on the rung below it anyway, so the two readings do not diverge in practice —
happy to reorder if you meant otherwise.

Implements the direction from our session on 2026-07-25.

### DIFF
--- a/.claude/scripts/work-priority-ladder.test.sh
+++ b/.claude/scripts/work-priority-ladder.test.sh
@@ -1,0 +1,97 @@
+#!/usr/bin/env bash
+#
+# Guards the work-selection ladder (maintainer direction 2026-07-25): a run picks work top-down —
+# live breakage, then EVERY open PR it owns or trusts INCLUDING ITS OWN DRAFTS, then security
+# issues, then bugs, then the oldest actionable issue.
+#
+# Why this needs enforcing rather than merely stating: the contract already said "PRs before issues"
+# and already said "stop starting, start finishing", and the pile still happened. The mechanism was a
+# SCOPING hole, not a missing rule — priority-1 was defined over `non-draft` PRs in Merge policy, so
+# an own draft matched no rung and was reachable only through a paragraph 1,800 lines away that read
+# as a precondition on opening new drafts. Measured 2026-07-25: 99 open own PRs, 100% drafts, none
+# ever promoted, median age 6.9 days, 18 already CLEAN and idle a median 5.3 days, 16 conflicted,
+# 49 of 88 untouched in the 24h after opening.
+#
+# So this guards the two properties that actually close that hole, plus the drift that reopened it:
+#   1. the ladder exists, is ordered, and names all five rungs;
+#   2. rung 1 explicitly covers OWN DRAFTS, and Merge policy's `non-draft` is explicitly scoped to
+#      the merge command rather than the sweep — the exact misreading that produced the pile;
+#   3. severity outranks age, so a Security issue is not queued behind an older Docs one;
+#   4. the run-loop skill agrees with the contract — three surfaces restate this ordering, and a
+#      silent divergence between them is how the previous wording drifted.
+
+set -euo pipefail
+
+repo_root="$(cd "$(dirname "${BASH_SOURCE[0]}")/../.." && pwd)"
+constitution="${repo_root}/AGENTS.md"
+maintenance_skill="${repo_root}/.claude/skills/portfolio-maintenance/SKILL.md"
+workflow="${repo_root}/.github/workflows/ci.yaml"
+
+fail() {
+  echo "work-priority ladder: FAIL — $*" >&2
+  exit 1
+}
+
+# Markdown prose is hard-wrapped, so a guarded sentence routinely spans two lines and exists on NO
+# single line. Flatten once and match substrings against the flattened copy; keep `grep -Fq` only
+# for things that genuinely live on one line (a heading, a table row, an identifier).
+constitution_flat="$(tr '\n' ' ' < "${constitution}" | tr -s '[:space:]' ' ')"
+skill_flat="$(tr '\n' ' ' < "${maintenance_skill}" | tr -s '[:space:]' ' ')"
+
+assert_prose() {
+  case "$2" in
+    *"$1"*) ;;
+    *) fail "$3" ;;
+  esac
+}
+
+# ── 1. the ladder exists and is ordered ──────────────────────────────────────
+grep -Fq '### The work-selection ladder — one ordering, checked top-down every run' "${constitution}" ||
+  fail "contract does not define the work-selection ladder"
+assert_prose 'you do not descend while a higher rung still has actionable work' \
+  "${constitution_flat}" "ladder does not state that rungs are strictly ordered"
+
+for rung in \
+  '| **0** | **Live breakage** |' \
+  '| **1** | **Open PRs — INCLUDING your own drafts** |' \
+  '| **2** | **Security issues** |' \
+  '| **3** | **Bugs** |' \
+  '| **4** | **Oldest actionable issue** |'; do
+  grep -Fq "${rung}" "${constitution}" ||
+    fail "ladder is missing rung row ${rung}"
+done
+
+# ── 2. rung 1 covers own drafts, and `non-draft` is scoped to the merge command ──
+assert_prose 'Rung 1 includes your own DRAFTS' \
+  "${constitution_flat}" "ladder does not put own drafts in rung 1 — the exact hole that produced the pile"
+assert_prose 'draft and non-draft alike' \
+  "${constitution_flat}" "rung 1 does not state it covers drafts and non-drafts alike"
+assert_prose 'scoping below bounds the merge COMMAND, never the SWEEP' \
+  "${constitution_flat}" "Merge policy does not scope its non-draft clause to the merge command"
+
+# ── 3. severity outranks age ──────────────────────────────────────────────────
+assert_prose 'Severity outranks age at rungs 2–3; age decides only *within* a rung' \
+  "${constitution_flat}" "contract does not state that severity outranks age"
+
+# ── 4. the run-loop skill agrees with the contract ────────────────────────────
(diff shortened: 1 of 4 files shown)


devantler:

devantler:



@coderabbitai review

devantler:

CodeRabbit (lane 1) is quota-limited at this head — `Review limit reached`, next review in ~36 min. That is a provider-quota state, not a verdict on this change. Walking the ladder to Codex rather than waiting on the clock.

devantler:



@codex review

devantler:

Codex (lane 2) is credit-exhausted at this head — `reached your Codex usage limits for code reviews`, no retry window stated, needs maintainer credits. Provider-quota state, not a verdict. Advancing to Cursor Bugbot (lane 3); its trigger follows as a bare comment, which is the one trigger that must carry no inline disclosure.

devantler:
@cursor review

AGENTS.md
```diff
@@ -452,18 +453,55 @@ governs the issue work that follows.) Two rules enforce that:
...
+Maintainer direction 2026-07-25: *"focus on open PRs before claiming new work … open prs > security
+issues > bugs > oldest issue. We want to stop starting and start finishing."* This ladder is the
+**single normative statement** of what a run picks up; every other ordering sentence in this contract
+defers to it. Rungs are strictly ordered — **you do not descend while a higher rung still has
+actionable work**:
+
+| # | Rung | What it covers |
+|---|---|---|
+| **0** | **Live breakage** | CI red on `main`, a broken build or site, an urgent security fix. Preempts everything and is the one exception to capture-before-you-build. |
+| **1** | **Open PRs — INCLUDING your own drafts** | Every actionable own/trusted PR in your lane, **draft and non-draft alike**, driven to a terminal state: merged, or parked on a **named, live-verified** blocker. Automation-owned dependency PRs are excluded (see *Merge policy*). |
```
devantler, line 469: nit (accepted consequence, not a defect): with 99 open own PRs, a strict "never descend while a higher rung has actionable work" means rungs 2-4 are not reached until the pile drains — roughly 4 days at the measured ~27 merges/day. That is the maintainer's stated intent ("focus on open PRs before claiming new work"), so I am recording it rather than softening it. The escape valve is already in the rung text: a PR parked on a named, live-verified blocker is *not* actionable, so rung 1 empties even while some PRs stay open. Flagging because a future reader may mistake the starvation for an oversight.

AGENTS.md
```diff
@@ -452,18 +453,55 @@ governs the issue work that follows.) Two rules enforce that:
...
+| **2** | **Security issues** | `type:"Security"`, regardless of age. |
```
devantler, line 470: nit: this rung depends on the org-level Issue Type `Security` existing and being applied. Verified live at this head — 18 open `type:"Security"` and 99 `type:"Bug"` across the org — so the rung is not vacuous today. But an issue that *is* a security problem and carries no Type silently lands in rung 4 instead of rung 2. The contract already mandates a Type on every issue (`Every issue carries an Issue Type — no exceptions`), so the failure mode is untyped-issue drift, not this rung; no change requested.

.claude/scripts/work-priority-ladder.test.sh
```diff
@@ -0,0 +1,97 @@
...
+  exit 1
+}
+
+# Markdown prose is hard-wrapped, so a guarded sentence routinely spans two lines and exists on NO
+# single line. Flatten once and match substrings against the flattened copy; keep `grep -Fq` only
+# for things that genuinely live on one line (a heading, a table row, an identifier).
+constitution_flat="$(tr '\n' ' ' < "${constitution}" | tr -s '[:space:]' ' ')"
+skill_flat="$(tr '\n' ' ' < "${maintenance_skill}" | tr -s '[:space:]' ' ')"
+
+assert_prose() {
```
devantler, line 41: Reviewed specifically for vacuity, since a guard that silently matches everything is worse than no guard. Two properties checked at this head:

1. **Literal matching.** Several guarded strings contain `*` (e.g. `only *within* a rung`). `case` patterns glob, so an unquoted expansion would match loosely and the guard could pass on mutated text. Verified the quoted `"$1"` inside the pattern is literal: the true string matches, and a probe replacing `*within*` with `XwithinX` does **not**. No false-positive path.
2. **Every guard RED-proven.** 9 negative controls, each asserting its probe actually mutated the file before trusting the result — the failure mode recorded in `prose-guards-must-flatten-whitespace`, where a probe that silently matched nothing produced a false "NOT CAUGHT".

No finding; recording the verification so the next reader need not re-derive it.

devantler (commented):

**Reviewed commit:** `2cb00407f7b2` (equal to current head)

All three external lanes were tried in contract priority order and none will deliver a review at this
head. Every one is a **provider spend/quota state, not a verdict on this change**:

| Lane | Tried | Outcome |
|---|---|---|
| CodeRabbit (1) | 17:41Z | `Review limit reached` — next review in ~36 min; account at the 95th percentile of review volume |
| Codex (2) | 17:43Z | `reached your Codex usage limits for code reviews` — **no window stated**; needs maintainer credits |
| Cursor Bugbot (3) | 17:45Z | check-run `neutral` + `output.title: Error`, 7s, zero inline comments → **the run never happened**; `Bugbot couldn't run - usage limit reached`, needs an admin to raise the Cursor spend limit |

Two of the three state **no retry window at all**, which is the condition the 2026-07-21 widening
covers — a provider's billing plan is not a judgement about our change, and a finished PR does not
park on it.

### Findings

Three were found during this review and **all three are already fixed in this head**, so they are
listed for the record rather than as open threads:

- **P1 — severity sorting amplifies cross-lane claim collisions.** *Claim protocol* opened by quoting
  the very selection rule this PR replaces, and the replacement makes its core hazard worse: sorting
  rungs 2–3 by severity narrows the target pool from "the oldest of ~368 open issues" to "one of 18
  open `type:"Security"` issues", so every instance aims at the same handful — while cross-lane
  arbitration is the known-broken hole in rule 4 (#2302). Fixed in `2cb0040`, scoped to rungs 2–3
  because rung 1 pulls the other way (an own draft belongs to exactly one lane).
- **nit — dangling cross-reference.** New-finds pointer used a bare *Capture before you build*; this
  contract's style for a sub-item is *Section → Sub-item*. Fixed in `223d536`.
- **nit — stale run floor.** The floor still named the oldest actionable issue as the ideal artifact,
  which the ladder had just superseded. Fixed in `223d536`.

Inline comments below record two accepted consequences and the vacuity verification of the new
enforcement test. None requests a change.

### Honest limitations of this review

A self-review is the weakest form of this gate and I am the least independent reader of my own diff.
Two scoping decisions a bot lane might have challenged: three incidental "oldest-first queue" phrases
in `product-engineering/SKILL.md` were left untouched (they describe *not jumping the queue*, which
survives the change), and rung-1 starvation of rungs 2–4 is accepted as the maintainer's stated
intent rather than mitigated.

**Verdict: no P0/P1 findings outstanding at `2cb00407f7b2`** (1 P1 and 2 nits found and fixed within
this PR).

devantler:

**Tried and evaluated as a user** (readiness condition 3) — exercised the ladder against live portfolio state at this head, rather than tracing it:

**Rung 1, old vs new selection.** Ran the old rung-1 predicate (actionable trusted-author, *non-draft*, excluding automation-owned) over all 172 open portfolio PRs: it matches **1 PR portfolio-wide, 0 of them own**. The new predicate matches **99 own drafts** — 18 `CLEAN` and promotable immediately, 16 `DIRTY` needing conflict work. So a run following this text has 34 immediately-actionable rung-1 items where it previously had none.

**Rungs 2–3, tiering.** Confirmed the Issue Types the rungs name are real and populated: **18 open `type:"Security"`, 99 `type:"Bug"`** among **368** open issues. Under the old age-only sort those 18 competed on age against the whole backlog; under the ladder they precede it.

**Enforcement exercised, not assumed.** All 9 negative controls run against a sandboxed copy, each asserting its probe actually mutated the file before trusting the result — every guard fires. Verified `assert_prose` matches literally despite `*` in guarded strings (a `case` glob would have made guards vacuous): true string matches, `XwithinX` probe does not. Confirmed the new CI job actually fired on this PR rather than being filtered out — `Test work-priority ladder` passed in 4s.

**Not exercised:** whether the ladder changes real run behaviour over time. That is hypothesis 17, baselined at 99 open own PRs / 18 `CLEAN` / 16 `DIRTY` / 6.9d median, `check_after: 2026-07-26`.

PR #2466, opened 2026-07-25
fix(review): retire the two-phase review reservation

devantler:

## Why

Every review request was posting an extra comment that renders as nothing but the disclosure line — the maintainer flagged it as recurring noise that costs tokens and GitHub rate limit. It was meant to reserve a review lane so two agent instances could not request the same provider at once, but measurement shows it never did that: the reservation and its own trigger went up 1–2 seconds apart from the same run, far too fast for another instance to ever see it.

Across a 7-day portfolio corpus it prevented **zero** collisions in 75 attempts, while costing 90 blank comments and double the write rate per review request on a rate-limited API.

## What

Removes the reservation step and keeps the request marker, now carried in the same comment as the trigger. Overlapping instances still see a live request and still avoid duplicating it — at half the comments. The retirement is recorded with its evidence so a future run does not reinstate it, and five contract guards were proven to fail without the change.

**This is a loosening of a coordination control.** What is removed: a pre-trigger lane reservation with an "oldest wins" election. What still covers the risk: the retained request marker plus the existing rules that one provider request may be active at a time and that markers are re-read immediately before triggering. The residual exposure is a duplicate review request if two instances trigger within seconds of each other — no such event occurred in the measured window, and the failure mode is one wasted provider call, not a correctness or safety failure.

Maintainer direction, interactive session 2026-07-25.


devantler:

Requesting review at `5a7904dd` — CI green (16 passed, 0 failed). This request deliberately follows the single-comment discipline this PR introduces: the request marker rides in the same comment as the trigger, with no separate pre-trigger reservation comment. Exercising the changed procedure on its own PR is the user-evaluation step.



@coderabbitai review

devantler:

CodeRabbit (lane 1) completed without a gate-satisfying artifact at `5a7904dd` — `Review limit reached`, next included review in 53 minutes (19:39:31Z). Its `✅ Action performed — Review finished` ack was, as always, not evidence a review ran. Advancing to lane 2 rather than waiting; still one provider request at a time.



@codex review

devantler:

Readiness record for `5a7904dd9c` — all three conditions hold, so promoting.

**1. Programmatically tested.** All 13 contract suites pass; CI green (16 passed, 0 failed). The five new guards in `review-provider-loop-contract.test.sh` were RED-proven individually, each with a verified-landed ablation (the marker was confirmed to have actually changed before reading the test result — a no-op ablation would otherwise look exactly like a robust guard).

**2. Reviewed.** Codex green at head: `Didn't find any major issues`, **Reviewed commit** `5a7904dd9c` == `headRefOid[0:10]`, 19:43:03Z. Lane 1 (CodeRabbit) completed without a gate-satisfying artifact at this head (`Review limit reached`, 53-minute window) and its `Review rate limited` commit status is excluded from mergeability by contract. Stopping at the first success — Bugbot not spent.

**3. Tried and evaluated as a user.** The change alters how a review is requested, so I exercised the new procedure on this PR rather than tracing it: both requests here used the single-comment form this PR introduces — request marker, prose and trigger in one comment, no pre-trigger reservation. **CodeRabbit parsed and acted on it** (acked, then reported its quota) and **Codex parsed it, reacted 👀, and delivered a full review**. Both lanes fire correctly at half the comments, which is the behavioural claim the change rests on.

One incidental confirmation: the upstream `agentic-engineering` plugin carries no reservation text, so this was a consumer-local rule and needs no upstream follow-up.

PR #2475, opened 2026-07-26
docs(agents): point skill authoring at agent-skills, not the synced plugin copy

### LINKED ISSUE #2476: AGENTS.md names a synced plugin SKILL.md as an authoring surface, so edits there are silently reverted

## Evidence

`AGENTS.md` § Agent definition locations lists
`plugins/agentic-engineering/skills/agent-improvement/SKILL.md` among the
version-controlled surfaces the Agent Improver may change.

Verified at the `agent-plugins` merged tip (2026-07-26): that file carries
`metadata.github-repo: https://github.com/devantler-tech/agent-skills` and
`github-ref: refs/tags/v1.9.1`. It is installed by `gh skill install` and re-pulled by
the daily `update-agent-skills` workflow. Its sibling
`plugins/agentic-engineering/agents/agent-improver.agent.md` carries **no** provenance
and is genuinely authored there — so editability is a per-file property, not a
per-directory one.

The same mistake was caught live on `agent-plugins#89` (2026-07-25) by CodeRabbit and
rated Major: a role rename had been applied to three bundled copies whose upstream is
`devantler-tech/agent-skills`.

## Problem / audience

Any agent following the contract routes a generic skill change into a file that accepts
it, passes CI, merges — and then loses it at the next sync, with no conflict, no failing
check, and no signal. The work is silently undone and the loss is attributed to nothing.

## Expected behaviour

The section names `devantler-tech/agent-skills` as the authoring home for generic skill
bodies, keeps the `agents/*.agent.md` half unchanged, and states the per-file provenance
check (`grep -l github-repo <path>`) so the next file added to a plugin is classified
rather than guessed.

## Acceptance criteria

- [ ] § Agent definition locations names no concrete `plugins/*/skills/*/SKILL.md` path
      as an authoring surface.
- [ ] It names `devantler-tech/agent-skills` as the upstream for skill bodies.
- [ ] It still names `plugins/agentic-engineering/agents/agent-improver.agent.md`.
- [ ] It states the `metadata.github-repo` per-file check.
- [ ] A guard test pins all four, is RED-proven per assertion, and runs in CI via a
      paths filter that also lists the guard itself and `ci.yaml`.

Rough size: XS.

devantler:

## Why

The contract told a future agent to edit a file that silently discards its edits.
`§ Agent definition locations` named the Agent Improver's skill body inside
`agent-plugins` as a place to author changes — but that copy is pulled in from another
repository every day, so an edit there looks fine, passes review, merges, and then
quietly disappears at the next sync. No conflict, no failing check, nothing to notice.

The trap is easy to fall into because the same folder holds both kinds of file: the
agent definition next to it really is authored there. This was caught for real on an
`agent-plugins` PR last week and rated a major finding.

## What

Points skill changes at the repository that actually owns them, leaves the agent half
alone, and states the one-line check that tells the two apart — so the next file added
to that folder gets classified rather than guessed at.

Adds a guard so the contract cannot drift back, wired into CI in a way that still runs
if a later change tries to delete the guard itself.

Fixes #2476

### DIFF
--- a/.claude/scripts/authoring-surface-provenance.test.sh
+++ b/.claude/scripts/authoring-surface-provenance.test.sh
@@ -1,0 +1,68 @@
+#!/usr/bin/env bash
+#
+# Guards § Agent definition locations against naming a SYNCED file as an authoring
+# surface.
+#
+# `plugins/*/skills/*/SKILL.md` in devantler-tech/agent-plugins carry
+# `metadata.github-repo` provenance and are re-pulled by the daily
+# `update-agent-skills` workflow. A hand-edit there survives until the next sync and
+# then silently disappears — no conflict, no CI failure, no signal. The sibling
+# `plugins/*/agents/*.agent.md` carry no provenance and ARE authored there, so
+# "editable here?" is a per-file question and a per-directory rule gets it wrong.
+#
+# The contract named the synced agent-improvement SKILL.md as a version-controlled
+# authoring target (caught 2026-07-25 on agent-plugins#89, rated Major), which would
+# route a generic skill change into a file that reverts it. This pins that shut.
+
+set -euo pipefail
+
+repo_root="$(cd "$(dirname "${BASH_SOURCE[0]}")/../.." && pwd)"
+constitution="${repo_root}/AGENTS.md"
+
+fail() {
+  echo "authoring-surface provenance: FAIL — $*" >&2
+  exit 1
+}
+
+[ -f "${constitution}" ] || fail "missing ${constitution}"
+
+# Flatten so a rule wrapped across lines is still matched (prose guards must
+# flatten whitespace — a line-anchored grep silently passes on reflowed text).
+flat="$(tr '\n' ' ' <"${constitution}" | tr -s ' ')"
+
+section="$(printf '%s' "${flat}" |
+  sed -n 's/.*### Agent definition locations\(.*\)### Authority model.*/\1/p')"
+[ -n "${section}" ] || fail "could not isolate § Agent definition locations"
+
+# 1. No CONCRETE bundled skill path may be named as an authoring surface. The glob
+#    form `plugins/*/skills/*/SKILL.md` is the warning itself and must stay allowed,
+#    so match only path segments that name a real plugin and a real skill.
+if printf '%s' "${section}" |
+  grep -qE 'plugins/[A-Za-z0-9_.-]+/skills/[A-Za-z0-9_.-]+/SKILL\.md'; then
+  fail "§ Agent definition locations names a concrete bundled
+  plugins/<plugin>/skills/<skill>/SKILL.md path as an authoring surface.
+  Those files are SYNCED from their metadata.github-repo upstream and an edit there is
+  silently reverted. Name the upstream authoring repo instead (agent-skills)."
+fi
+
+# 2. The true upstream for skill bodies must be named, by its FULL slug. A bare
+#    `agent-skills` would be satisfied by the `update-agent-skills` workflow name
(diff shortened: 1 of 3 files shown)


devantler:

## Readiness evidence @ `2868c47ca2`

**Programmatically tested.** The guard has 4 assertions, each **individually RED-proven**
by mutating the fixed `AGENTS.md` and confirming that assertion — and only it — fires:

| # | Assertion | Mutation | Result |
|---|---|---|---|
| 1 | no concrete `plugins/<p>/skills/<s>/SKILL.md` named | re-insert the old path | FAIL ✓ |
| 2 | `devantler-tech/agent-skills` named | rewrite the slug | FAIL ✓ |
| 3 | `agents/agent-improver.agent.md` still named | rename it | FAIL ✓ |
| 4 | `metadata.github-repo` check stated | drop the key | FAIL ✓ |

Unmutated: 4/4 pass. Full-tree RED/GREEN also run against the pre-fix `AGENTS.md` from
`origin/main` (FAIL) and the fixed text (OK).

Two of these assertions were **vacuous when first written** and were caught by the probes,
not by inspection — `agent-skills` was satisfied by the `update-agent-skills` workflow
name in the same paragraph, and a bare `github-repo` by neighbouring prose. Both were
tightened to the full slug and the exact metadata key.

Full contract suite: **15/15 green**. Repo CI at this head: **25 success / 7 skipped / 0 failure**.

**Wiring verified behaviourally, not by inspection.** The job's paths-filter output was
initially *not declared* on the `changes` job, which would have left
`needs.changes.outputs.authoring-surface` empty and the job silently never running —
a green CI that proves nothing. Declared it, then cross-checked that every filter has an
output and every output has a consumer (both clean), and confirmed the check-run actually
executed at this head rather than skipping:

```
Test authoring-surface provenance — success (00:21:39Z → 00:21:44Z)
```

**Tried as a user.** I used the new instruction on its own subject this run: ran
`grep -l github-repo` against the merged `agent-plugins` tip and it correctly classified
`skills/agent-improvement/SKILL.md` as synced (`github-repo: …/agent-skills`,
`github-ref: refs/tags/v1.9.1`) and `agents/agent-improver.agent.md` as authored
(no provenance). Confirmed `agent-skills` really does own `agent-improvement/SKILL.md`.
The instruction answers the question it claims to answer.

devantler:


@coderabbitai review

devantler:

Closing as a duplicate of #2474, which predates this by 78 minutes (23:02:48Z vs
00:21:21Z). I lost that race and did not detect it: my pre-flight claim scan covered
`platform` and `ksail` but never ran against `monorepo`, so an open PR in my own lane
was invisible to me.

Per the claim protocol I abandon rather than compete. I diffed the two implementations
first, and **#2474 is the better one on every axis I checked** — no finding to carry over:

- It extends the existing `agent-role-delivery-contract.test.sh`, whose paths filter
  already lists `AGENTS.md`, the script, and `ci.yaml`. That is correct wiring with **no**
  workflow change at all; this PR added a new job, filter and output for the same coverage.
- It scopes its assertions to a **single bullet** and says explicitly that asserting
  against the whole contract is a scope hole. I hit exactly that hole here: two of my four
  assertions were **vacuous** on first write — `agent-skills` was satisfied by the
  `update-agent-skills` workflow name in the same paragraph, and a bare `github-repo` by
  neighbouring prose.
- It resolves provenance **structurally**, `yq '.metadata.github-repo'` at the exact YAML
  path, instead of my line grep — which is precisely what makes the two failure modes above
  unrepresentable rather than merely fixed.

Issue #2476 (filed here for the defect) stays open and is the tracking issue for #2474.

PR #2493, opened 2026-07-26
fix(scripts): fail submodule-init when init leaves the submodule empty

### LINKED ISSUE #2492: submodule-init.sh reports 'isolated ✓' and exit 0 for an EMPTY submodule — fail-open on its primary job

## Evidence

Reproduced deterministically 2026-07-26 in a second superproject worktree while the same submodule was already checked out in a sibling worktree:

```
$ bash .claude/scripts/submodule-init.sh libraries/agent-plugins
Submodule path 'libraries/agent-plugins': checked out 'a96f3c80...'
submodule-init: libraries/agent-plugins — isolated ✓
$ echo $?
0
$ ls -A libraries/agent-plugins | wc -l
0
```

Re-running is idempotently wrong: `isolated ✓`, exit 0, **zero entries**, every time.

`git submodule update --init` reported success and exited 0 while populating nothing. Any subsequent `git -C libraries/agent-plugins …` then walks **up to the superproject** — `rev-parse --show-toplevel` returns the worktree root and `remote -v` returns the *monorepo* URL, which is how this was noticed.

## Problem

`init_repair_probe` takes the empty branch, runs `git submodule update --init "$path"`, then `repair`, then `probe` — and **nothing asserts the population actually happened**. `probe` verifies *isolation*, not *content*, so it passes vacuously on an empty directory.

The script's own header calls itself "init + repair + probe (fail-closed)". It is fail-closed on isolation and **fail-open on its primary job**.

`is_populated()` already exists and is used correctly to *skip* unpopulated submodules in `--check` (an uninitialised submodule is legitimately empty there). Init mode is the opposite case: it was **asked** to populate, so an empty tree afterwards is a failure, not a skip.

## Why this matters beyond a confusing error

An empty submodule makes globs over it match **nothing**, so checks that read submodule content pass vacuously rather than failing loudly. `AGENTS.md` § *Agent definition locations* mandates exactly such a glob to decide whether a bundled skill is safe to edit:

```sh
for f in libraries/agent-plugins/plugins/*/skills/*/SKILL.md; do ... done
```

Against an empty submodule that loop iterates zero times and reports nothing — a clean bill of health on the ownership check that gates whether editing a file is safe. This is the same fail-open class the script's header already records having fixed twice before.

## Smallest useful change

In init mode only, assert the post-condition: after `git submodule update --init`, `die` unless `is_populated` holds. `--check` behaviour is unchanged.

## Acceptance criteria

- [ ] Init mode exits non-zero with a clear message when the submodule is still empty afterwards
- [ ] `--check` still skips legitimately-uninitialised submodules
- [ ] RED-proven: a test fails before the fix and passes after

devantler:

## Why

The helper that prepares a submodule for editing could report success — `isolated ✓`, exit code 0 — while leaving the submodule **completely empty**. It did this repeatedly and identically on every re-run, so there was no signal anything was wrong.

That is worse than an outright error. An empty submodule makes searches over it match **nothing**, so checks that read its contents come back clean instead of failing. One of those is the check our own conventions require before editing a file that is synced in from another repository — so the safety check that decides "is this file safe to edit?" could quietly return a clean answer having examined nothing at all.

## What

The helper now verifies the thing it was actually asked to do: if the submodule is still empty after being set up, it stops with a clear message instead of reporting success. Checking an already-empty, deliberately-uninstalled submodule is unaffected — that case is legitimate and still passes.

One caveat stated plainly: the original false-success needed conditions that could not be recreated in a test, so the test proves the guard fires and says openly which of its assertions do not depend on the fix.

Fixes #2492

### DIFF
--- a/.claude/scripts/submodule-init.sh
+++ b/.claude/scripts/submodule-init.sh
@@ -269,6 +269,17 @@ init_repair_probe() {
     # Fresh (empty) submodule: populate it at its pinned commit, then relocate the `core.worktree`
     # that `git submodule update` writes into the shared config.
     git submodule update --init "$path"
+    # It can exit 0 having populated NOTHING — observed 2026-07-26 running from a linked superproject
+    # worktree while a sibling worktree already held that submodule: git printed `checked out '<sha>'`,
+    # exited 0, and left the directory empty. `probe` below verifies ISOLATION, not content, so it
+    # passes vacuously on an empty directory and the script reports `isolated ✓` for a submodule that
+    # was never checked out. That is fail-open on this script's PRIMARY job, and it is worse than a
+    # loud failure: globs into an empty submodule match nothing, so content checks — including the
+    # bundled-skill ownership audit AGENTS.md mandates before editing a synced file — pass vacuously
+    # instead of erroring. Assert the post-condition: init mode was ASKED to populate, so an empty
+    # tree is a FAILURE here, never the legitimate skip `--check` makes of an uninitialised submodule.
+    is_populated "$path" ||
+      die "'$path' is STILL EMPTY after 'git submodule update --init' (which exited 0) — do not read or edit it"
     repair "$path"
   fi
   probe "$path" || die "repair did not restore isolation for '$path' — do not edit it"
--- a/.claude/scripts/submodule-init.test.sh
+++ b/.claude/scripts/submodule-init.test.sh
@@ -322,6 +322,56 @@ report "scoped cleanup: a pre-existing sibling worktree still exists afterwards"
 report "scoped cleanup: that sibling worktree is still FUNCTIONAL (not just present)" \
   "$(git -C "$c10_sib" rev-parse --show-toplevel >/dev/null 2>&1 && echo yes || echo no)"
 
+# 11. #2492 — init mode must FAIL when `git submodule update --init` exits 0 without populating.
+#     Observed running from a linked superproject worktree while a sibling worktree already held the
+#     submodule: git printed `checked out '<sha>'`, exited 0, left the directory EMPTY, and the script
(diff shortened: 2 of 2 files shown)


devantler:

@coderabbitai review

devantler:

@coderabbitai review

devantler:

## Readiness evaluation @ `e12bf9a5aa0da9ea857855dc715bb703715ef397`

**1. Programmatically tested.** Suite run locally: **24/24 pass**. RED/GREEN ablation performed — removed only the `is_populated` post-condition and its `die`, then re-ran:

- `empty-init: the failure NAMES the empty submodule` → **FAIL**, with exactly the message the PR predicts: `submodule-init: sub — not checked out here; nothing to probe`.
- The two assertions the PR **labels non-discriminating** did indeed still pass with the guard ablated, and `--check still SKIPS a legitimately deinitialised submodule` was unaffected.

The honest limit written into the test file is accurate as written — verified rather than taken on trust.

**2. Reviewed.** CodeRabbit clean at this head (`No actionable comments were generated`); the only body section is the excluded `🔇 Additional comments (2)`. Zero unresolved threads, `CLEAN`, all required checks green.

**3. Tried and evaluated as a user.** Ran the built script against a real, uninitialised submodule from a linked superproject worktree — the exact execution model this guard is about:

- populated `libraries/agent-skills` at its pinned commit,
- reported `submodule-init: libraries/agent-skills — isolated ✓`, exit 0,
- `git -C libraries/agent-skills rev-parse --show-toplevel` returned the **worktree's own path**, not a `.git/modules/...` gitdir.

The check that mattered most as a user: **the new fail-closed post-condition does not false-positive on a legitimate init.** A guard that turned every real populate into a failure would be worse than the fail-open it replaces, and it does not.

Promoting.

PR #2499, opened 2026-07-27
fix(surveyor): stop prescribing gh --json fields the subcommand rejects

### LINKED ISSUE #2498: Surveyor overlay prescribes gh --json fields the target subcommand rejects (52 errors in 7 days)

## Evidence

The surveyor overlay `.claude/agents/portfolio-surveyor.md` prescribes `gh --json` field
lists containing fields the target `gh` subcommand does not accept. Every surveyor run that
follows those lines emits `Unknown JSON field` and has to retry.

Measured over the portfolio session corpus, 2026-07-20 → 07-27 (131 session files, Claude lane;
the Codex lane records no error flag, so this is a floor, not a total):

| Field | Prescribed on | Occurrences |
|---|---|---|
| `reviewThreads` | `gh pr view --json` | 24 |
| `merged` | `gh pr view --json` | 16 |
| `headRefName` | `gh search prs --json` | 4 |
| `mergedAt` | `gh search prs --json` | 2 |
| `headRepository` | `gh search prs --json` | 2 |
| `authorAssociation` | `gh pr view --json` | 2 |
| `baseRepository` | `gh pr view --json` | 2 |
| | **total** | **52** |

`Unknown JSON field` is the largest non-Edit tool-error signature in the window.

## Root cause

`gh pr view` and `gh search prs` expose **disjoint** `--json` vocabularies — 47 fields vs 19:

- `pr view`-only (invalid on search): `mergedAt`, `headRefName`, `headRefOid`, `mergeStateStatus`,
  `reviewDecision`, `statusCheckRollup`, …
- search-only (invalid on `pr view`): `authorAssociation`, `commentsCount`, `repository`, …
- in **neither** (GraphQL-only names): `reviewThreads`, `merged`, `baseRepository`.

Three sites in the overlay cross that boundary. All three reproduce:

1. **line 88** — the core per-PR hygiene-pentad command, the most-executed instruction in the
   deployment — carries `reviewThreads` in a `gh pr view --json` list.
2. **line 330** — `gh pr view <n> --json comments,reviewThreads`.
3. **line 326** — `gh search prs … --json number,repository,mergedAt`. This one is also
   *unimplementable as written*: the surrounding prose insists on keying the window on `mergedAt`
   "never `updatedAt`", but `mergedAt` does not exist on the search surface.

Thread state was never obtainable from `--json`: the file already documents the correct
GraphQL form at line 130 (`reviewThreads(first:100, after:$cursor){nodes{isResolved} …}`), so
the invalid field is redundant as well as wrong.

## Scope

The defect is **local to this consumer overlay**. The upstream plugin's
`portfolio-surveyor.agent.md` does not mention `reviewThreads` at all, so no upstream hop is
needed. The overlay's other five prescribed field lists (lines 43, 45, 183, 360, 361) were each
validated live and are correct.

## Expected behaviour

Every `gh … --json` field list in the overlay is accepted by the subcommand it is written for,
and the merge-window intent at line 326 is preserved without the unavailable field.

## Acceptance criteria

- [ ] `reviewThreads` removed from both `gh pr view --json` lists; thread state sourced from the
      GraphQL query the file already documents.
- [ ] Line 326 keeps its merge-time window (the `--merged-at` search qualifier already enforces it
      server-side) without requesting `mergedAt` from the search surface.
- [ ] A contract check RED-proves each site and fails if a cross-surface field is reintroduced.
- [ ] No other prescribed field list regresses.

Size: S.

## Attribution correction (added after the initial evidence pass)

The 52 occurrences are **not** all caused by the three prescribed sites. Splitting them by whether
the overlay actually told the agent to use the field:

| Route | Fields | Occurrences |
|---|---|---|
| **Prescribed by the overlay** | `reviewThreads` (24), `mergedAt` on search (2) | **26** |
| **Improvised** — same confusion, not prescribed anywhere | `merged` (16), `headRefName` (4), `headRepository` (2), `authorAssociation` (2), `baseRepository` (2) | **26** |

Verified by extracting every `gh … --json` list from the pre-fix overlay: `merged`,
`authorAssociation` and `baseRepository` appear nowhere in it, and `headRefName`/`headRepository`
are prescribed only on `gh pr view`, where they are valid.

So removing the three bad prescriptions fixes **half** the signature outright. The other half is the
orchestrator reaching for GraphQL/`pr view` names on a search surface unprompted — which is why the
fix also states the vocabulary split explicitly at the point of use, rather than only deleting the
offending fields. That half is expected to fall, not to vanish.

devantler:

## Why

The surveyor's own instructions told it to ask GitHub for data GitHub does not offer on that
command, so part of the every-run PR health check failed and had to be retried. One of the three
broken instructions was self-contradictory: it insisted on keying a search on a value that search
cannot return.

This class of error happened 52 times in the past week — the largest avoidable error signature we
have. Half of those trace to these three instructions; the other half is the same confusion reached
by hand, so the fix also states the rule at the point of use rather than only deleting the bad
fields. Expect that half to fall, not to vanish.

## What

Corrects the three instructions to ask the right surface for each piece of data, keeping the
original intent intact — review-thread state already had a documented correct route, and the
merge-time window is enforced by the search qualifier itself. Adds a check that fails the build if a
future edit reintroduces the same class of mistake.

No behaviour change for the products; this is the agent's own definition.

Fixes #2498

### DIFF
--- a/.claude/agents/portfolio-surveyor.md
+++ b/.claude/agents/portfolio-surveyor.md
@@ -85,8 +85,11 @@ public and private — no per-repo loop needed to enumerate):
    `cursor[bot]` — **exact login match, never a substring**; `cursor[bot]` is trusted here only on the PR-author surface;
    `Copilot`/`copilot-swe-agent[bot]` are NOT trusted), pull the
    heavy fields one PR at a time:
-   `gh pr view <n> --repo devantler-tech/<repo> --json number,state,mergeStateStatus,reviewDecision,statusCheckRollup,mergedAt,reviewThreads,headRefName,headRefOid,author,body,files`
-   — do **not** pull `statusCheckRollup` for every PR in every repo. When the current-head pentad is
+   `gh pr view <n> --repo devantler-tech/<repo> --json number,state,mergeStateStatus,reviewDecision,statusCheckRollup,mergedAt,headRefName,headRefOid,author,body,files`
+   — do **not** pull `statusCheckRollup` for every PR in every repo.
+   ⚠️ **Thread state is NOT available here.** `reviewThreads` is a GraphQL-only field, so requesting
+   it from `gh pr view` fails with `Unknown JSON field`. Get (b) from the paginated GraphQL query
+   below, never by adding that field to this list (monorepo#2498). When the current-head pentad is
    clear (CLEAN + required checks + zero threads/body findings + a current-head green
    review), classify trusted-bot **non-drafts** as **MERGE-READY** and trusted-bot **drafts** as
    **REVIEW-READY**; otherwise **NEEDS-FIX** and name the gate. A `devantler` PR always follows the
@@ -323,11 +326,16 @@ public and private — no per-repo loop needed to enumerate):
      post-merge PR comment is a primary steering channel, and an open-PR-only sweep would never
      surface it — so in addition to every open `devantler` PR, sweep the PRs **merged in the last
      ~3 days** (bounded: `gh search prs --owner devantler-tech --author devantler --merged
-     --merged-at ">=<UTC date 3 days ago>" --limit 100 --json number,repository,mergedAt` — key the
-     window on `mergedAt`, never `updatedAt`, which post-merge edits can inflate) for the
+     --merged-at ">=<UTC date 3 days ago>" --limit 100 --json number,repository,url` — the
+     `--merged-at` **qualifier** is what keys the window on merge time, so the sweep never depends on
+     `updatedAt`, which post-merge edits can inflate. ⚠️ Do **not** request `mergedAt` in this
+     `--json` list: the search surface exposes only 19 fields and `mergedAt` is not among them, so it
+     fails with `Unknown JSON field`. Read the timestamp per-PR from `gh pr view --json mergedAt` if
+     a run actually needs the value — monorepo#2498) for the
      same candidate-comment signal. For each such PR — **including drafts** — also
-     pull `comments` and the review-thread replies:
-     `gh pr view <n> --repo devantler-tech/<repo> --json comments,reviewThreads`. **Apply the disclosure
+     pull `comments`, then the review-thread replies via the paginated GraphQL query above
+     (`reviewThreads` is GraphQL-only, so `gh pr view` cannot return it):
+     `gh pr view <n> --repo devantler-tech/<repo> --json comments`. **Apply the disclosure
      disambiguator before flagging** (the same one the PR-ownership rule above uses, per the contract's
      *Untrusted input → Distinguish the human maintainer from yourself*): the agent also comments as
      `devantler`, so a bare exact-login match is NOT enough. A `devantler` comment whose body carries a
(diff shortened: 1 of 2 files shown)


devantler:

@coderabbitai review

devantler:

## Readiness record

**1. Programmatically tested.** RED-proved before the fix (guard exits 1, naming the pentad command),
GREEN after. **Ablated in both directions** so the guard is known to discriminate rather than merely
pass: reintroducing `reviewThreads` on a `gh pr view` list → RED; reintroducing `mergedAt` on the
search list → RED; restored → GREEN. Added an extractor-breakage assertion so a regex that stops
matching fails loudly instead of passing vacuously. All six sibling contract tests re-run green;
`shellcheck` clean across the added block (the SC2016 infos elsewhere are pre-existing). CI green at
this head, including `Test portfolio surveyor contract`.

**2. Reviewed.** CodeRabbit, current head `4d5a190`, no actionable findings — three `🔇 Additional`
notes, all `LGTM!`. Zero review threads, zero inline comments, no service/rate-limit marker.

**3. Tried and evaluated as a user.** Every corrected command was executed against real PRs, not
reasoned about:

| Command | Result |
|---|---|
| corrected pentad `gh pr view --json …` (site 1) | returns the PR, e.g. `{"n":2491,"s":"MERGED","head":"claude/agents-draft-intake-cap-2490"}` |
| corrected merged-window `gh search prs --json number,repository,url` (site 3) | returns 3 merged PRs inside the `--merged-at` window |
| corrected `gh pr view --json comments` (site 2) | returns 6 comments |
| the documented GraphQL route for thread state | `threads=2 unresolved=0` |

Each of the three original forms was also run first and confirmed to fail with `Unknown JSON field`,
so the defect and the fix were both observed, not inferred.

**Note on the evidence.** The initial write-up implied these three sites explained all 52 errors.
Re-attribution against the pre-fix file showed **26 prescribed / 26 improvised**; the issue and PR
body were corrected before this review. The improvised half is expected to fall, not vanish.

devantler:

## Evidence correction — the headline count was inflated, and the emphasis was wrong

A record-level recount after merge showed **every** transcript line carrying `Unknown JSON field`
carries the string **more than once**, so the raw occurrence figure (52) counts repetitions within
records, not distinct failures. It is an upper bound. Re-measured at session granularity over the
same corpus (131 portfolio session files, 07-20 → 07-27, excluding this improver session's own
diagnostics):

| Field | Route | Distinct sessions affected |
|---|---|---|
| `merged` | **improvised** | **8** |
| `reviewThreads` | prescribed | 3 |
| `headRefName` | improvised | 3 |
| `headRepository` | improvised | 2 |
| `authorAssociation` | improvised | 2 |
| `baseRepository` | improvised | 2 |
| `mergedAt` | prescribed | 1 |

**15 of 131 sessions** (11.5%) hit the class at least once.

**What this changes.** The occurrence count over-weighted `reviewThreads` — 24 raw hits, but
concentrated in only 3 sessions. By breadth the dominant signature is `merged`, which this overlay
never prescribed. So the three corrected prescriptions are a smaller share of the real-world problem
than the merged PR body claims.

**What it does not change.** The three prescriptions were invalid and were reproduced failing live
before and after the fix, independent of any count — that reproduction, not the frequency figure, is
what justified the change. It also raises the relative value of the part of this PR that states the
`pr view` / `search` / GraphQL vocabulary split **at the point of use**: that guidance is what
addresses `merged`, the widest offender, which deleting fields could never have reached.

Recorded in the improver's durable memory, and hypothesis 22's baseline has been restated on the
session-granularity metric so the follow-up is actually scoreable.

Issue #2526, opened 2026-07-28
Portfolio map has drifted from the live org: a missing product and three stale visibility labels


devantler:
> 🤖 Generated by the Agentic Engineer

## Evidence

Surfaced by the portfolio survey on 2026-07-28 and verified live against the org API (not from a cached list):

| Repo | `AGENTS.md` says | Live |
|---|---|---|
| `doggy-countdown` | *absent from the map entirely* | public, not archived, created 2026-07-26 |
| `wedding-app` | "(private)" | **public** |
| `ascoachingogvaner` | "(private)" | **public** |
| `unifi` | "(private)" in the map body | **public** |

## Why it matters

Two concrete consequences, not cosmetics:

1. **Board eligibility is decided by visibility.** *Every issue belongs on the board* makes a private repo's issue a maintainer decision, because project 5 is public. Three repos the map labels private are in fact public, so their open issues are board-eligible and are almost certainly unboarded — a coverage gap the board's own product card treats as a defect.
2. **A product exists with no entry anywhere.** `doggy-countdown` is in neither the Portfolio map nor the Stack map, so it has no `AGENTS.md` row, no product card, no owning-repo row for intake, and no Epic or `roadmap` issue. Nothing routes work to it and no run will ever select it.

The contract already anticipated this class — it tells a run to determine visibility live "never from a hard-coded list or from the portfolio map's parenthetical (both go stale)", citing `unifi` in that exact state on 2026-07-18. That guidance stopped a run acting on stale data; it did not fix the stale data, and two more repos have since drifted the same way.

## Acceptance criteria

- The Portfolio map lists `doggy-countdown` with a path and per-repo `AGENTS.md` link, or records explicitly why it is out of scope.
- The Stack map gains a row for it (what it is good for + owning repo), or the omission is stated.
- The three "(private)" parentheticals are corrected to match live visibility.
- Board coverage is backfilled for whichever of the three now-public repos have unboarded open issues (`board-add.sh` is idempotent and refuses private repos, so it is safe to run across them).
- Ideally a check that fails when the map and the live org set diverge, so this does not silently drift a fourth time.

Rough size: S for the map edits, M with the backfill and a drift check.

devantler:
> 🤖 Generated by the Agentic Engineer

Independently re-derived this on the 13:0xZ survey and can add one row to the count, plus a consequence the "why it matters" section does not yet name.

[...]

PR #2529, opened 2026-07-28
docs(agents): stop recording repository visibility in the portfolio map

devantler:

## Why

The portfolio map labelled the wedding app and the AS Coaching site "private". Both are public, and have been for a while. That mislabel is not cosmetic: whether a repo is private decides whether its issues may go on the public project board, so a row that wrongly says "private" quietly suppresses tracking that should be happening.

It also keeps recurring — the board section already warns that this exact label goes stale and cites a previous instance, and two more repos drifted the same way since.

## What

Corrects both rows (they also gained the per-repo `AGENTS.md` link the map claimed they did not have), and states at the map itself that visibility is deliberately not recorded there — look it up live instead. Removing the field is what stops a third recurrence; correcting the two values alone would not.

Scoped to the visibility half of #2526 so it stays one concern. Still open there: adding `doggy-countdown` to both maps or recording why it is out of scope, and the optional drift check. Evidence for both is on the issue. Board backfill appears already satisfied — this run's survey enumerated the board in full and found no public open issue missing.

Part of #2526

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -32,12 +32,19 @@ keeping them healthy *and* moving them forward.
 | UniFi Crossplane provider (shared lib) | `devantler-tech/provider-upjet-unifi` | `libraries/provider-upjet-unifi` | [AGENTS.md](https://github.com/devantler-tech/provider-upjet-unifi/blob/main/AGENTS.md) |
 | Kyverno policy library (shared lib) | `devantler-tech/kyverno-policies` | `libraries/kyverno-policies` | [AGENTS.md](https://github.com/devantler-tech/kyverno-policies/blob/main/AGENTS.md) |
 | World at Ruin (game) | `devantler-tech/world-at-ruin` | `applications/world-at-ruin` | [AGENTS.md](https://github.com/devantler-tech/world-at-ruin/blob/main/AGENTS.md) |
-| Wedding app (private) | `devantler-tech/wedding-app` | `applications/wedding-app` | (private) |
-| AS Coaching (private) | `devantler-tech/ascoachingogvaner` | `applications/ascoachingogvaner` | (private) |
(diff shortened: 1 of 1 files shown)


devantler:

CI is green at this head (17 success, 0 failures). CodeRabbit is inside a stated per-developer review-limit window measured at 16:09Z with ~49 minutes remaining, so it will not deliver here yet.




@codex review

devantler:

## Readiness at `25bb419212ba716cf15be8c6b2741bc0ad399813`

**1 — Programmatically tested.** Every `.claude/scripts/*.test.sh` contract test run locally against this branch; all pass. (`agent-role-delivery-contract` fails in a fresh worktree only because `libraries/agent-plugins` is uninitialised there — it passes in an initialised checkout, and CI is authoritative.) CI at this head: **17 success, 11 skipped, 0 failures, 0 pending.**

**2 — Reviewed.** `codex@25bb419212` — *"Didn't find any major issues"*, reviewed commit matching this head. CodeRabbit was inside a stated per-developer review-limit window measured at 16:09Z, recorded as `cr:no-gate` above, so the lane loop advanced.

**3 — Tried and evaluated as a user.** This changes prose that agents read, so the check is whether its claims are true against live state rather than a runtime behaviour:

- `wedding-app` and `ascoachingogvaner` both return `private=false` from the org API — the two rows this corrects.
- Both have an `AGENTS.md` at their default branch, so the links the rows now carry resolve rather than 404 (the old cells said `(private)` and linked nothing).
- No `(private)` marker remains anywhere in the Portfolio map. The one surviving `UniFi network (private)` string is inside *Every issue belongs on the board*, quoted deliberately as the precedent for this failure mode — left intact on purpose.

Judged as its reader: a run consulting the map for board eligibility now finds no visibility claim at all and is pointed at the live query, which is the only source that cannot go stale between edits.

PR #2551, opened 2026-07-29
fix(scripts): name the resolved slug in branch-cleanup's abort message

### LINKED ISSUE #2550: branch-cleanup's slug guard prints two identical strings when it rejects, and AGENTS.md does not say the owner is implied

## Evidence — two failed invocations of a mandated every-run step, 2026-07-29

End-of-tick branch hygiene is a duty on every run. Reaching it cost two wasted attempts because the guard that rejects a bad slug cannot be acted on from its own message.

`AGENTS.md:2297` documents the call as:

```
branch-cleanup.sh <repo_path> <slug> <manifest> [apply|dry-run] [namespace]
```

`<slug>` is undefined there. Read alongside the surrounding prose — which uses "slug" throughout for the per-session worktree slug — the natural first reading is the session slug. That fails:

```
$ branch-cleanup.sh . clever-carson-5cd023 <manifest> dry-run claude
clever-carson-5cd023: ABORT — slug 'clever-carson-5cd023' does not match the checkout at '.', whose
  origin is 'devantler-tech/monorepo'.
```

Fair enough — the message names the origin, so the obvious correction is to pass that origin. **That also fails, and this is the actual defect:**

```
$ branch-cleanup.sh . devantler-tech/monorepo <manifest> dry-run claude
devantler-tech/monorepo: ABORT — slug 'devantler-tech/monorepo' does not match the checkout at '.', whose
  origin is 'devantler-tech/monorepo'.
```

The message asserts that two **byte-identical strings** do not match. Nothing in it is actionable, and the natural next step is to doubt the checkout rather than the argument.

## Root cause

`branch-cleanup.sh:176` prepends the owner itself:

```sh
slug_lc=$(printf 'devantler-tech/%s' "$SLUG" | tr '[:upper:]' '[:lower:]')
```

So the parameter is the **bare repository name** (`monorepo`), and passing `devantler-tech/monorepo` builds `devantler-tech/devantler-tech/monorepo` internally. The comparison is correct; the message is not, because it reports the raw `$SLUG` rather than the derived value it actually compared.

The guard itself is good and should stay fail-closed — an unverified slug would fetch the keep-set for the wrong repository and a wrong keep-set deletes open PRs' branches. This is only about being able to act on its rejection.

## Impact

Low severity, high recurrence: it is on the path of a step every run of every instance is required to perform, and the misleading form is reached precisely by following the first message's own hint. An instance that concludes the checkout is broken rather than the argument may skip the sweep entirely, which is how stale branches accumulate silently.

## Acceptance criteria

- The rejection message reports the value actually compared, so the two sides are never printed as identical strings — e.g. name the derived `devantler-tech/<slug>` and state that the owner is implied.
- The message says what to pass (the bare repository name), not only that the argument was wrong.
- `AGENTS.md:2297` names the parameter unambiguously (`<repo-name>`), so it cannot read as the session worktree slug.
- A test covers the owner-qualified input (`devantler-tech/monorepo`) rejecting with an actionable message, since that is the input the first error steers a caller toward.

Rough size: XS. Both sweeps otherwise behaved correctly once invoked properly — `claude` and `cursor` namespaces each reported 0 deletions, with no-PR branches listed as candidates rather than removed.

devantler:

## Why

The end-of-tick branch sweep refuses to run when its slug argument disagrees with the checkout's origin — a good guard, since a wrong slug would fetch the keep-set for the wrong repository and delete branches that belong to open PRs. But the rejection could not be acted on. The parameter is the bare repository name and the script prepends the owner itself, while the message printed the raw argument beside the parsed origin. Pass the owner-qualified form — which is exactly what the first rejection's own hint steers you toward — and it asserts that two byte-identical strings do not match.

I hit this during this run's own cleanup: two failed invocations of a step every run is required to perform, and I only resolved it by reading the source. The risk is an instance concluding its checkout is broken rather than its argument, and skipping the sweep — which is how stale branches accumulate quietly.

## What

Report the value actually compared and say what to pass; name the parameter `<repo-name>` in the contract so it cannot read as the session worktree slug. The guard's decision is untouched — it refuses exactly the same inputs as before; only the explanation changes.

Fixes #2550

### DIFF
--- a/.claude/scripts/branch-cleanup.sh
+++ b/.claude/scripts/branch-cleanup.sh
@@ -187,8 +187,13 @@ check_remote_identity() {
   _nwo=$(github_nwo "$_url")
   if [ -n "$_nwo" ]; then
     if [ "$_nwo" != "$slug_lc" ]; then
-      echo "$SLUG: ABORT — slug '$SLUG' does not match the checkout at '$REPO_PATH', whose" >&2
-      echo "  $_kind is '$_nwo'. The keep-set would be fetched for the wrong repository." >&2
+      # Name the value actually COMPARED ($slug_lc), not just the raw argument.
+      # <slug> is a bare repo name and the owner is prepended, so an
+      # owner-qualified argument prints an identical-looking pair otherwise:
+      #   slug 'devantler-tech/x' ... whose origin is 'devantler-tech/x'
+      echo "$SLUG: ABORT — slug '$SLUG' (resolved to '$slug_lc') does not match the checkout" >&2
+      echo "  at '$REPO_PATH', whose $_kind is '$_nwo'. The keep-set would be fetched for the" >&2
+      echo "  wrong repository. <slug> is a BARE repo name — 'devantler-tech/' is prepended." >&2
       exit 2
     fi
   elif [ "$MODE" = "apply" ] && [ "${BRANCH_CLEANUP_ALLOW_UNVERIFIABLE_ORIGIN:-}" != "1" ]; then
--- a/.claude/scripts/branch-cleanup.test.sh
+++ b/.claude/scripts/branch-cleanup.test.sh
@@ -192,6 +192,24 @@ report "slug that disagrees with origin is refused (exit 2)" \
 report "slug refusal names both the slug and the origin repo" \
   "$(grep -q 'ksail' <<<"$out" && grep -q 'does-not-exist-2531' <<<"$out" && echo yes || echo no)" "out=$out"
 
+# --- 7a. an OWNER-QUALIFIED slug must be diagnosable ----------------------
+# <slug> is a BARE repo name; the helper prepends `devantler-tech/`. Passing the
+# owner-qualified form — the shape `gh -R` takes everywhere, so the likely
+# mistake — resolves to `devantler-tech/devantler-tech/<repo>` and is correctly
+# refused. But the message printed the RAW argument next to the parsed origin,
+# so for a slug whose repo half is right it rendered as a contradiction:
+#   slug 'devantler-tech/x' does not match the checkout ... whose origin is 'devantler-tech/x'
+# Identical strings, no stated reason — the reader cannot see that a second
+# owner was prepended. The refusal must name the value actually COMPARED.
+git -C "$slugwork" remote set-url origin "https://github.com/devantler-tech/does-not-exist-2531.git"
+manifest="$tmp/manifest7a"
+: >"$manifest"
+out="$("$helper" "$slugwork" "devantler-tech/does-not-exist-2531" "$manifest" dry-run claude 2>&1)" && rc=0 || rc=$?
+report "owner-qualified slug is refused (exit 2)" \
+  "$([[ ${rc:-0} -eq 2 ]] && echo yes || echo no)" "rc=${rc:-0} out=$out"
+report "owner-qualified refusal names the RESOLVED slug, not just the raw argument" \
+  "$(grep -q 'devantler-tech/devantler-tech/does-not-exist-2531' <<<"$out" && echo yes || echo no)" "out=$out"
+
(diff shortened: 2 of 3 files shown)


devantler:



CI green at this head (13 success / 18 skipped, `CLEAN`).

Self-review before asking: `shellcheck` clean on both files; the full `branch-cleanup.test.sh` contract passes; the new assertion was proved **RED** on the parent commit and **GREEN** here. I also ran the real helper against this checkout both ways — the owner-qualified slug now prints the resolved value and the cause, and the correct bare slug still completes its sweep unchanged (`keep 419`), so the guard's decision is untouched.

Scope for the reviewer: this changes only a diagnostic message. The thing worth checking is that it cannot leak a *wrong* impression in the other direction — that no input now reads as accepted when it is refused, and that `$slug_lc` is always the value actually compared at that point.

@coderabbitai review

devantler:



CodeRabbit returned **Review limit reached** rather than a review — a per-account quota state, not a verdict on this change. Under the contract's rate/billing-limited rule that counts as the lane not delivering, so the ladder advances to Codex rather than waiting the quoted window out; a sibling run measured today that the quoted window itself grows under adaptive limiting (13min → 51min), so planning around it is not reliable.

Note this head is `40c85e77`, not the `08c4400b` the earlier request named — the AGENTS.md half was pushed after it, so that request was stale regardless.

Requesting Codex once CI settles on this head; I will not spend a review on an incomplete build.

devantler:



CI settled clean at this head — 0 failures, `mergeStateStatus: CLEAN`. CodeRabbit is rate-limited (recorded above), so the ladder advances here.

What I already verified, so you can spend attention elsewhere: the new assertion is RED on the parent commit and GREEN here; `shellcheck` is clean on both scripts; and I ran the real helper against this checkout in both directions — the owner-qualified slug now prints the resolved value and the cause, and the correct bare slug still completes its sweep unchanged (`keep 419`).

The one thing worth your scrutiny: this touches a **fail-closed destructive guard**. The change should be message-only. Please check that no input which was previously refused can now proceed, and that `$slug_lc` is genuinely the value compared at that point rather than one that could drift from it later.

@codex review

devantler:

## Readiness — self-promoting

**Green review at head:** `codex@40c85e7776` — *"Didn't find any major issues"*, `Reviewed commit: 40c85e7776`, matching this PR's head exactly. CodeRabbit was rate-limited at this head (`cr:no-gate` recorded above), so the ladder advanced to Codex, which served.

**Programmatically tested:** 20 checks success / 11 skipped, 0 failures, `CLEAN`, 0 unresolved threads, 0 review-body findings.

**Tried and evaluated as a user** — this change has a real runtime surface, so I exercised it rather than reasoning about it:

| Probe | Result |
|---|---|
| `branch-cleanup.sh . devantler-tech/monorepo … dry-run claude` (the invocation that cost me two attempts) | now prints `slug 'devantler-tech/monorepo' (resolved to 'devantler-tech/devantler-tech/monorepo')` and names the bare-repo-name requirement |
| `branch-cleanup.sh . monorepo … dry-run claude` (positive control) | completes unchanged — `keep 419`, 0 deleted |
| New owner-qualified assertion | **RED** on the parent commit, **GREEN** here |
| Full `branch-cleanup.test.sh` contract | all assertions pass |
| `shellcheck` on both scripts | clean |

The positive control is the one that matters here: this is a fail-closed destructive guard, so the thing to prove is that a message change did not alter which inputs are refused. It did not — the same sweep runs, with the same keep-set, and only the rejection text differs.

One correction for the record: I twice reported here that Codex had not responded. It had — at 06:54:48Z, four minutes before my first check. My probes filtered on login `chatgpt-codex-connector`, but the actual login carries the `[bot]` suffix, so every one of them was structurally blind to a review that was already sitting on the PR. A control probe (group-by over all comment authors) is what surfaced it.

PR #2587, opened 2026-07-30
fix(agents): count Codex comment-form findings so the pentad stops reading clear over an open P2

### LINKED ISSUE #2577: bug(ai-engineer): a Codex P2 posted as an issue comment leaves the hygiene pentad reading fully clear

## Evidence (measured on monorepo#2559, head `948bb06f73`, 2026-07-29 ~20:2xZ)

At one head, Codex posted **two issue comments 41 seconds apart**:

| Time | Surface | Content |
|---|---|---|
| 19:44:35Z | issue comment | `## Review finding` — **P2 — Fail closed when probing supported config paths** |
| 19:45:16Z | issue comment | `Codex Review: Didn't find any major issues. :+1:` + `**Reviewed commit:** 948bb06f73` |

Codex counts only **P0/P1** as "major", so the green and an open P2 coexist at the same head.

At that head Codex produced **no review object and no inline thread**:

- `pulls/2559/reviews` — Codex review objects exist only at `3d940b6f13` (16:14:04Z) and `5b0f8ab5c1` (19:07:45Z). **None at `948bb06f73`.**
- `reviewThreads` — 12 total, **all resolved**, first comments at 16:14:05Z / 18:39:43Z / 19:07:45Z. **None at `948bb06f73`.**
- newest CodeRabbit review object is at `5b0f8ab5c1`, i.e. **stale** relative to the head.
- CI at head: **22 success / 8 skipped / 0 failing**; `mergeable=true`, `mergeStateStatus=CLEAN`.

## The failure

Every hygiene-pentad item reads clear while a real P2 is open:

| Pentad item | Reads | Actual |
|---|---|---|
| (a) checks green | ✅ 22/8/0 | correct |
| (b) unresolved threads | ✅ 0 | correct — the finding never became a thread |
| (c) non-thread review findings | ✅ none | **WRONG** — procedure inspects CodeRabbit review bodies; the newest CR review is stale, so nothing is counted |
| (d) conflict | ✅ CLEAN | correct |
| (e) green review at head | ✅ `codex@948bb06f73` | correct — and it is the *same* surface carrying the finding |

So a run following the documented procedure literally promotes and merges a PR with an open, valid P2. The finding here was a genuine **fail-open in a drift guard** (`findConfigs` swallowing every `os.Stat` error for config paths), so the cost is shipping a guard that reports success when it could not read the file.

## Root cause — and a correction to #2244

`AGENTS.md`'s green-review table gives the Codex lane a *findings* artifact of "review object, `state: COMMENTED`, inline threads", and #2244 states it more strongly:

> Only its *findings-bearing* results are review objects.

**That premise is false.** It held at this PR's three earlier heads and did not hold at `948bb06f73`. Codex can publish **both** its green **and** its findings as issue comments, so a procedure that reads review objects + threads + CodeRabbit bodies is structurally blind to a comment-form finding.

This is the same blind-spot class `AGENTS.md` already documents for Bugbot ("a gate or survey that sweeps only `pulls/<n>/reviews` and `issues/<n>/comments` is structurally blind to it") — recurring for the Codex lane, in the more dangerous direction.

**#2244 and this issue are the two halves of one root cause**, with opposite costs:

- **#2244** — green-in-comment missed ⇒ false `green_review=none` ⇒ an unwarranted self-review. Cost: a wasted review.
- **this issue** — finding-in-comment missed ⇒ false "pentad clear" ⇒ promote + merge with an open finding. Cost: shipping the defect.

## Expected behaviour

Item (c) must sweep the **Codex issue-comment surface** for the current head, not only CodeRabbit review bodies and threads. Concretely: a `chatgpt-codex-connector[bot]` issue comment whose body contains a `## Review finding` section is a **non-thread review finding** and blocks promotion until fixed-or-refuted with a disclosed resolution reply — exactly as a CodeRabbit body finding does.

Note the ordering trap: the green comment can be **newer** than the finding comment (41s later here), so "latest Codex comment wins" is not a safe rule. `Didn't find any major issues` must not clear a P2 that has no resolution record.

## Acceptance criteria

1. Item (c) counts `## Review finding` sections in current-head `chatgpt-codex-connector[bot]` **issue comments**, in addition to the existing CodeRabbit body sections.
2. A newer Codex green comment does **not** clear an older same-head Codex comment finding lacking an authenticated disclosed resolution reply.
3. `AGENTS.md`'s Codex row records that **both** green and findings may be comment-form; the "only findings are review objects" premise is corrected there and on #2244.
4. A test fixture reproduces the exact monorepo#2559 shape — green comment newer than an unresolved P2 comment at the same head, zero threads, stale CR review — and the pentad evaluates **not clear**.
5. Ablation: removing the new comment-surface sweep turns criterion 4 red and nothing else.

**Size:** small-to-medium. Detection is a comment-surface filter; the care is in the ordering rule (2) and in not double-counting a finding that *did* arrive as a thread.

Part of the same review-lane reliability theme as #2244, #2561 and #2468.

devantler:

## Why

The readiness gate could pass a pull request that still had an open, valid review finding — so a run following the documented procedure would promote and merge work a reviewer had objected to.

Codex reports problems in two different places, and we were only looking in one of them. On monorepo#2559 it raised a P2 as an ordinary comment, then posted its "no major issues" note 41 seconds later on the same commit (it only counts the most severe two levels as "major"). Every one of the five readiness checks reported clear while the objection was still open and unanswered.

## What

The readiness gate now also reads Codex comment-form findings, so an open one blocks promotion until it is fixed or refuted on the record — exactly like every other reviewer finding. A later "no major issues" note no longer clears an earlier objection on the same commit, and a finding that cannot be tied to a specific commit is treated as current rather than ignored.

Guarded by a contract test built from the real monorepo#2559 comments, so if Codex changes how it reports, the test fails rather than the gate silently going blind again.

Fixes #2577


devantler:

## Self-review before requesting a reviewer

Ran the pre-submission pass over my own diff. Recording the one real trade-off I found, rather than leaving a reviewer to discover it.

**Fail-closed attribution can over-report.** An unattributable `## Review finding` comment counts as current-head until a disclosed resolution reply clears it. If Codex ever posts a finding with no code citation, an old finding from a superseded head would read as open until someone replies to it.

I kept it fail-closed deliberately. This is a safety gate, and the failure directions are not symmetric: over-reporting costs one resolution reply, while under-reporting is exactly the defect being fixed — a merged PR with an unanswered P2. Every finding comment observed so far cites code, so attribution should nearly always succeed. Flagging it because a reviewer may weigh that differently.

**Verification performed**
- RED/GREEN: the seven new assertions fail on current `main` content and pass after the change.
- Ablation, 7 arms: each guarded phrase removed individually makes the test fail *with its own message*; each arm asserts the file's sha actually changed first, so a no-op edit cannot read as a pass. Files restored byte-identically.
- Behavioural ablation, 2 arms: corrupting the fixture's permalink sha and deleting the fixture each fire their own assertion.
- Sibling contract tests re-run; `agent-role-delivery-contract` fails identically on unmodified `origin/main` content (uninitialised submodule in a fresh checkout), so it is pre-existing and not from this change.

devantler:

@coderabbitai review

devantler:

## Tried and evaluated as a user — against live production data

Ran both the old and the new rule against the real monorepo#2559 at the reported head `948bb06f73`, rather than only against the fixture. This is the reported failure reproduced end to end, and then closed.

**Old rule — item (c) sweeps CodeRabbit review bodies + threads only**
- newest CodeRabbit review sits at `5b0f8ab5c1`, i.e. **stale** relative to the head → contributes 0
- unresolved threads: **0**
- **item (c) = 0 → reads CLEAR**, while a valid P2 is open and unanswered

**New rule — also counts Codex `## Review finding` issue comments, attributed by `/blob/<sha40>/`**
- 1 finding section found, attributed to `948bb06f73809c3aee43501ed81a0904d0fa218e`
- that is **exactly** the head → counts toward the gate
- **item (c) = 1 → reads NOT CLEAR, promotion blocked**

So the change does the thing it claims on the case that motivated it: the gate that previously waved this PR through now stops it.

**Also exercised**
- Full contract suite green in CI (0 failing), including `Test review-provider loop contract`.
- Ablation, 9 arms, all guarded — 7 prose arms each producing their own distinct failure message, plus 2 behavioural arms on the fixture. Every arm verified the file sha actually changed first, so a no-op edit could not read as a pass.
- Files restored byte-identically after ablation.

devantler:

## Handoff state (2026-07-30T10:2xZ, head `e0799805`)

**This PR needs exactly one clean review at head, then promote + merge.** Everything else is done:

- CI **green, 0 failing** — including `Test review-provider loop contract`, `CodeQL`, `zizmor`, `CI - Required Checks`
- `mergeStateStatus: CLEAN`, 0 unresolved threads, 0 non-thread findings
- self-review posted (with the one trade-off I found), user evaluation posted (old vs new rule against live monorepo#2559)

**Review lane state:** CodeRabbit was requested at this head and returned `Review limit reached — next review available in 4 minutes`. That is a plan-quota refusal, not a verdict, so it does not gate merge. A retrigger is armed for after the stated window. If it still has not delivered on the next run, advance to Codex per lane priority — do **not** spend a second lane while the first is merely throttled.

**One deliberate scope line, so a later run does not "finish" it here:** the rule is prose in the surveyor overlay because that overlay is the authoritative surface today. It is generic reviewer-lane behaviour and belongs upstream in `agent-plugins` eventually — recorded as item 4b in `.claude/plugin-consumption/agentic-engineering-surveyor-diff.md` so removing the overlay cannot silently drop it. Porting it upstream is a separate change, not this one.

devantler:

@coderabbitai review

devantler:

## Correction to my own trade-off note — the evidence is weaker than I implied

In the self-review above I wrote that "every finding comment observed so far cites code, so attribution should nearly always succeed." I have now actually counted it, and the number does not support that phrasing.

Sampled the last 60 PRs in each of `monorepo`, `ksail`, `platform` and `world-at-ruin` (~240 PRs) for Codex `## Review finding` comments:

- comment-form findings found: **1**
- attributable via blob permalink: **1**
- unattributable: **0**

So attribution held on 1 of 1. That is not "nearly always" — it is a single observation, and it happens to be the very case that motivated the fix. I should not have characterised one data point as a general pattern.

Two things follow, and they pull in opposite directions:

1. **The over-report risk I flagged is unmeasured, not low.** With n=1 I cannot say how often a finding comment lacks permalinks. Fail-closed remains the right default for a safety gate, but the justification is the asymmetry of the failure directions, not frequency evidence.
2. **The shape is rare, which is exactly why it was missed.** One occurrence in ~240 PRs is far too infrequent to be caught by anyone eyeballing review output, and severe when it happens — the gate reports clear over an unanswered finding. Rarity is an argument for the mechanical guard, not against it.

Hypothesis 27 in my durable memory already tracks both directions, so if the fail-closed rule starts producing false NEEDS-FIX the next run will catch it and narrow the predicate rather than weaken the gate.

devantler:

CodeRabbit returned `Review limit reached` twice at this head (10:17Z, and again after its own stated 4-minute window). That is a plan-quota refusal rather than a verdict, so advancing to the next lane in priority order.

@codex review

devantler:

Pushed `d1198eb` — a coherence fix I found reviewing my own change: `body_findings`' sha was defined only as the CodeRabbit review's `commit_id`, so a run counting a Codex comment-form finding had no stated sha to emit. That push stales the request at `e079980`, so re-requesting at the new head.

Staying on Codex rather than restarting at CodeRabbit: CodeRabbit refused twice for plan quota within the last ten minutes, so its limit will not have reset. Recording that as a persisted no-gate rather than spending another cycle on a lane that is demonstrably throttled.

@codex review

devantler:

## Readiness — promoting

Pentad at `d1198eb6c6`, verified after the last push (not inherited from the earlier head):

| Item | State |
|---|---|
| (a) checks | 19 success / 12 skipped / **0 failing** |
| (b) unresolved threads | **0** |
| (c) non-thread findings | **0** — CodeRabbit body sections 0 (no review), **Codex comment-form findings 0** |
| (d) conflict | `CLEAN` |
| (e) green review at head | Codex — `Reviewed commit: d1198eb6c6` == head |

Item (c) was evaluated with the rule this PR adds, applied to this PR. That is the point: the check that was structurally blind on monorepo#2559 now runs on its own change, and reports zero because there genuinely is nothing there — not because it cannot see.

Three readiness conditions:
1. **Programmatically tested** — RED/GREEN proof; 9 ablation arms, each with a sha-changed precondition; full contract suite green in CI.
2. **Reviewed** — Codex green at the current head. CodeRabbit was tried first per lane priority and refused twice on plan quota, which is service state, not a verdict.
3. **Tried and evaluated as a user** — ran the old and new rules against live monorepo#2559: old yields item (c) = 0 (CLEAR, the fail-open), new yields 1 attributed to the exact head (NOT CLEAR).

PR #2627, opened 2026-08-02
refactor: rename gitops-tenant-template to platform-tenant-template

devantler:

## Why

GitOps is the *technology* the template uses, not the product it is a tenant of. The tenant belongs to the Platform, so `platform-tenant-template` names what it actually is. Maintainer direction, 2026-08-02.

## What

The repository has already been renamed on GitHub. This updates the monorepo's side of it: the submodule (path, section name and URL), both maps in `AGENTS.md`, the Cursor environment, the worktree-isolation runbook, the surveyor's repo list, and the documentation page.

Two judgement calls worth surfacing:

- The runbook's **dated findings are left verbatim** — only the paths they point at are updated, because those identify a submodule that still exists under a new name.
- The docs page rename **changes a public URL**, so `astro.config` gains a redirect from the old path. Bookmarks and the platform's own docs already link to it.

"GitOps" is kept where it genuinely describes the delivery mechanism, and dropped only where it was naming the tenant.

## Operational note

The submodule's section name and URL both change, so existing checkouts need `git submodule sync` on next pull. Nothing breaks without it — GitHub redirects the old URL — but the config will read stale.

Companion PR: devantler-tech/platform-tenant-template#126 (the template's own self-references, including the workflow guards that key on its literal repository name).


devantler:


@coderabbitai review

devantler:



**CodeRabbit recorded as no-gate at `7beb2a03` — advancing to Codex.** Its auto-generated summary for this head carries `Review limit reached`, so the review did not run; a quota notice is service state, not a verdict on this change.

CI is green at this head with 0 unresolved threads.


@codex review

devantler:

**Readiness — promoting and merging.**

- **Programmatically tested.** CI green at `7beb2a03` (23 checks, 0 failing, 0 pending), which includes the docs build and the Starlight links validator — the check that would catch the renamed page's inbound links.
- **Reviewed.** Codex green at exactly this head (`Reviewed commit: 7beb2a035f`), with 0 inline findings and no `## Review finding` comment. CodeRabbit was recorded `no-gate` here on a `Review limit reached` notice, which is service state rather than a verdict.
- **Exercised.** `.gitmodules` verified to carry the new path, section name and URL with 0 residual old references anywhere in the tree; the docs page rename is paired with an `astro.config` redirect so the previously published URL still resolves.

**Operational note for anyone with an existing checkout:** the submodule's section name and URL both changed, so `git submodule sync` is needed on the next pull. Nothing breaks without it — GitHub redirects the old URL — but the local config will read stale.

PR #2632, opened 2026-08-02
fix(scripts): report an uninitialised submodule apart from broken isolation

### LINKED ISSUE #2631: worktree-cleanup-all reports an uninitialised submodule as broken isolation

## Evidence

`worktree-cleanup-all.sh` reports two structurally different conditions with one message. Line 93
SKIPs any path whose `git rev-parse --show-toplevel` does not equal itself, and labels every such case
**"broken isolation"**:

```
### SKIP /…/monorepo/templates/platform-tenant-template (broken isolation: toplevel=/…/monorepo)
```

Observed in the live scheduled log (`~/.claude/worktree-cleanup-manifests/cleanup.log`, every 6h).
That path is **not** a broken-isolation case — it is an **uninitialised submodule**:

- the directory is empty (0 entries), and
- `git submodule status templates/platform-tenant-template` returns the `-` prefix (not initialised).

Control that discriminates the two, run at the same moment: `applications/ksail` returns its **own**
toplevel with `core.worktree` correctly set to itself, so the check does distinguish a healthy
submodule — the null is real, not an artifact of the instrument.

The script's own comment at line 79 already acknowledges the overlap ("A path that is not a repository
at all is handled by the toplevel check below"), so the conflation is known at the code level but not
reflected in what the operator is told.

The existing test bakes it in too: `t_skips_broken_isolation` does `rm -rf …/nested/.git` — which
constructs the *uninitialised* case — and then asserts the output says `broken isolation`. So there is
currently **no** test covering genuinely-broken worktree isolation.

Whoever reads that log — in practice the scheduled agent instances. The two conditions have opposite
remedies:

| Reported as | Actually | Correct remedy |
|---|---|---|
| broken isolation | uninitialised submodule | `submodule-init.sh <path>` — benign, expected |
| broken isolation | broken isolation (stray `core.worktree`) | repair in place; **sessions are silently colliding in one tree** |

Because the message is emitted every 6 hours it actively reinforces the wrong belief: this agent's own
durable memory carries `templates/platform-tenant-template` as a broken-isolation regression, sourced
from exactly this line. The cost is a run attempting an invasive `core.worktree` repair on a submodule
that is simply not checked out — or, in the other direction, reading a genuine collision as "just
uninitialised" and editing through an alias into the main checkout.

The SKIP itself is correct and fail-closed in both cases. **Only the diagnosis is wrong** — this is not
a safety regression, it is a wrong instrument reading.

## Expected behaviour

Distinguish the two before printing, using the presence of the path's own git metadata as the
discriminator (no `.git` entry ⇒ uninitialised/not-a-repo; own `.git` but mismatched toplevel ⇒ broken
isolation), and name the matching remedy in each message.

## Acceptance criteria

- [ ] An uninitialised submodule (or non-repository path) SKIPs with a message naming it as such and
      pointing at `submodule-init.sh`, not "broken isolation".
- [ ] A genuinely-broken-isolation path (own gitdir, `core.worktree` resolving outside itself) SKIPs
      with the existing "broken isolation" message.
- [ ] Both cases still SKIP and still never sweep through the alias — the fail-closed behaviour is
      unchanged, and a test pins that.
- [ ] `t_skips_broken_isolation` is corrected to assert the uninitialised label for the fixture it
      actually builds, and a **new** test covers real broken isolation (currently untested).

## Rough size

S — one branch in `worktree-cleanup-all.sh` plus two tests.

devantler:

## Why

The scheduled worktree sweep prints the same "broken isolation" message for two conditions that mean
opposite things. One is benign — a submodule that simply is not checked out. The other means live agent
sessions are silently editing the same physical tree and colliding.

Because the sweep runs every 6 hours, the wrong label gets reinforced continuously: it had already been
recorded as a genuine isolation regression in an agent's durable notes, sourced from this line. The cost
is a run attempting an invasive repair on a submodule that just needs initialising — or, worse, reading
a real collision as harmless.

## What

Tells the two apart and names the matching remedy in each message. The skip itself is unchanged and
still fail-closed in both cases — only the diagnosis changes.

Also corrects the existing test, which built the *uninitialised* fixture while asserting the
broken-isolation wording, and adds a test for genuinely-broken isolation, which nothing covered before.

Fixes #2631

### DIFF
--- a/.claude/scripts/worktree-cleanup-all.sh
+++ b/.claude/scripts/worktree-cleanup-all.sh
@@ -91,7 +91,20 @@ sweep() { # <repo_path>
     printf 'worktree-cleanup-all: ABORTING — cannot resolve physical path of %s\n' "$path" >&2
     exit 2; }
   if [ "$toplevel" != "$expected" ]; then
-    printf '\n### SKIP %s (broken isolation: toplevel=%s)\n' "$path" "$toplevel"
+    # Both conditions land here and both must SKIP, but they are NOT the same finding
+    # and their remedies are opposite, so they are reported apart. Discriminate on the
+    # path's OWN git metadata (a directory for a plain repo, a gitdir FILE for a
+    # submodule): absent means there is nothing checked out here and the toplevel
+    # merely resolved up to the parent; present means a real repository is aliased
+    # somewhere else, which is the dangerous case.
+    if [ ! -e "$path/.git" ]; then
+      printf '\n### SKIP %s (not initialised: no git metadata; toplevel resolved to %s)\n' \
+        "$path" "$toplevel"
+      printf '###      remedy: .claude/scripts/submodule-init.sh %s\n' "$path"
+    else
+      printf '\n### SKIP %s (broken isolation: toplevel=%s)\n' "$path" "$toplevel"
+      printf '###      remedy: repair this submodule in place — sessions editing here collide in one tree\n'
+    fi
     return 0
   fi
   # "$ROOT" is QUOTED: unquoted it is a glob pattern, so a root path containing
--- a/.claude/scripts/worktree-cleanup-all.test.sh
+++ b/.claude/scripts/worktree-cleanup-all.test.sh
@@ -86,15 +86,56 @@ t_rewrites_session_worktree_root() {
   rm -rf "$root"
 }
 
-t_skips_broken_isolation() {
-  # A submodule path whose toplevel resolves elsewhere must be SKIPPED, never swept
-  # through the alias (that would operate on the wrong tree entirely).
+t_skips_uninitialised_submodule() {
+  # An UNINITIALISED submodule (no git metadata of its own) resolves up to the parent
+  # repo and must be SKIPPED — never swept through that alias.
+  #
+  # It must NOT be reported as broken isolation. The two conditions have opposite
+  # remedies: this one is benign and cleared by submodule-init.sh, whereas broken
+  # isolation means live sessions are silently colliding in one physical tree. This
+  # fixture builds the uninitialised case (it deletes the metadata outright), so
+  # asserting the broken-isolation wording here is what let the two blur together.
   local root; root=$(make_root)
   rm -rf "$root/repo/nested/.git"          # now resolves up to the parent repo
   mkdir -p "$root/repo/nested/.claude/worktrees"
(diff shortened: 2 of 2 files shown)


devantler:

@coderabbitai review

CI is green here (12 success, 22 skipped, 0 failures) and the branch is CLEAN against base.

The change is a diagnostic split in the scheduled worktree sweep: one SKIP message covered both an uninitialised submodule and genuinely-broken worktree isolation, which have opposite remedies. Worth focusing on whether the `-e "$path/.git"` discriminator is the right one, since that is the whole load-bearing part.

devantler:
## Readiness evidence at `6e3f8139`

**Condition 1 — programmatically tested.** RED first: the new uninitialised-case test failed against
the unfixed script (10 passed, 1 failed). After the fix, 11/11.

The part worth stating is the **ablation**, because a passing test here proves less than it looks.
Both branches print a SKIP, so a discriminator that always took one path would still satisfy a naive
assertion. Inverting the discriminator (`[ ! -e "$path/.git" ]` → `[ -e … ]`) flipped **exactly the two
isolation tests** (9 passed, 2 failed) and restoring it returned 11/11 — so the branch is load-bearing
rather than decorative.

**Condition 2 — reviewed.** CodeRabbit at this head: substantive summary, no actionable finding
sections (only `🔇 Additional comments (2)`, which is its informational section), 0 unresolved threads.

**Condition 3 — tried and evaluated as a user.** This has a real runtime surface, so I exercised it
rather than reasoning about it: ran the actual script, in dry-run, against the actual monorepo, and
read the line for the actual submodule that prompted the issue.

Before:
```
### SKIP …/templates/platform-tenant-template (broken isolation: toplevel=…/monorepo)
```
After:
```
### SKIP …/templates/platform-tenant-template (not initialised: no git metadata; toplevel resolved to …/monorepo)
###      remedy: .claude/scripts/submodule-init.sh …/templates/platform-tenant-template
```

That submodule is empty and `git submodule status` reports it with the `-` prefix, so "not initialised"
is the correct reading and "broken isolation" was not. The control that makes this a real distinction
rather than a relabel: `applications/ksail` resolves to its own toplevel with `core.worktree` set, so
the underlying check does discriminate a healthy submodule.

CI: 12 success, 22 skipped, 0 failures; CLEAN against base. The one unresolved status context is
CodeRabbit's own, not a CI check.

Issue #2626, opened 2026-08-02
The agent's own self-measurement instruments have no Epic, and they keep producing the same class of defect


devantler:
> 🤖 Generated by the Agentic Engineer

## Why this Epic exists

The agent decides what to work on, whether a fix worked, and whether it is safe — using instruments it built for itself. Five defects in those instruments have now been filed as unrelated one-offs, and they are not unrelated: every one is a **measurement that counted something other than what it claimed to**, in a direction that quietly corrupts a decision.

| Issue | The instrument | What it actually counted |
|---|---|---|
| #2521 | injection detector | a runtime string, dominating the record count |
| #2610 (closed by #2612) | dispatch health | transcripts, not dispatches — 49 of 134 were subagent sidechains |
| #2621 | drift check | reports the stagger invariant UNKNOWN on every run, gated on a proof it does not need |
| #2622 | hypothesis scoring | the agent's own documentation of a signature, as occurrences of it |
| #2625 | reliability section | records from outside the window, because a resumed session rewrites file mtime |

No Epic covers them, so each is picked up by age against unrelated product work, the shared root cause is never addressed, and the next one arrives the same way.

## Audience and problem

The reader here is the agent itself and the maintainer reading its scorecard. The problem is that a wrong number in this layer does not look wrong — it looks like a finding. #2622 is the sharpest case: it would have reported a completely successful fix as a failure, arguing for redoing correct work. #2621's would have reverted a working cadence change.

## Hypothesis

These share a root cause worth fixing once rather than five times: **a count is defined by the thing it is named after, not by the population it actually enumerates.** A shared discipline — name the population, pair every count with a control that can distinguish it, ablate the detector before believing a zero — would have caught all five at authoring time.

## Success signal

Fewer instrument defects discovered *after* the instrument has been used to score something. Proxy: of the next instrument changes shipped, the share that arrive with a paired control and an ablation arm. Baseline: #2612 and #2624 both did; #2521 and #2621 predate the practice.

## Smallest useful first step

Adopt the existing children rather than build anything new: attach #2521, #2621, #2622 and #2625 to this Epic, then decide whether the discipline belongs in the `agent-improvement` skill upstream (generic, portable) or in this deployment's overlay.

## Acceptance criteria

- The five issues above are linked as sub-issues.
- A decision is recorded on where the shared discipline is authored, and it names the owning repository.
- Each remaining open child closes with an ablation arm proving its detector fires.

[...]

PR #2659, opened 2026-08-04
fix(telemetry): match PGP and RFC 7468 private-key labels

### LINKED ISSUE #2655: Private-key markers with RFC 7468 / PGP labels are matched by neither the redactor nor the detector

## Evidence

Every private-key marker pattern in `.claude/scripts/agent-telemetry.sh` — the awk redactor's four regexes, the three marker `sed` rules, `CRED_RE`, and `CRED_TABLE_RE` — spells the label as `[A-Z ]*PRIVATE KEY`. That misses two real label families:

- **RFC 7468 §2** defines `labelchar` as any printable US-ASCII character except hyphen-minus (`%x21-2C / %x2E-7E`), so digits and punctuation are legal in a label. `[A-Z ]*` matches neither.
- **OpenPGP ASCII armor (RFC 4880 / RFC 9580)** uses `-----BEGIN PGP PRIVATE KEY BLOCK-----`. The label does not *end* in `PRIVATE KEY`, so the pattern cannot match it however the character class is widened — the trailing ` BLOCK` has to be accommodated too. This is the format `gpg --armor --export-secret-keys` produces, so it is not hypothetical.

For an unmatched marker the redactor's pass A skips the opener, pass B never runs, and the key body is emitted unchanged.

## Why this is not fixed in #2654

It is **pre-existing on `main`** and untouched by that PR, which fixes a different defect (the terminated/unterminated classification). More importantly it cannot be fixed in the redactor alone: the file's own standing rule is that a shape the redactor masks but the detector misses reports **"clean"**, which is the worst failure mode a leak detector has, and the suite enforces that parity per shape. Widening the marker is therefore one coordinated change across the redactor, both detector regexes, and the parity tests — not a character-class edit.

## Acceptance criteria

- One widened label expression, applied to **every** marker site listed above in the same change, preserving the exact five-dash boundary.
- Coverage for the PGP armor label, whose label does not end in `PRIVATE KEY`.
- Coverage for a label containing digits and punctuation.
- Detector/redactor parity asserted for each new shape — both detected **and** redacted — so the widening cannot land on one side only.
- Real-corpus counts unchanged, both canaries at zero.

Rough size: S–M. Behaviour-preserving for every non-key line.

devantler:

## Why

The telemetry miner's job is to keep secrets out of the reports the agents read and write. It recognised private keys by a label pattern that could not spell two real formats — including the one `gpg` produces by default. A key in either format was written out in full **and** missed by the leak detector, so the report would have said "clean" over it.

## What

Widens the private-key label to cover both formats, applied at all nine places the old pattern appeared so the masker and the detector cannot drift apart again. Proven with before/after evidence on the real code path, plus checks in both directions that ordinary text and public certificates are still left alone.

No behaviour change for anything that is not a private key.

Fixes #2655

### DIFF
--- a/.claude/scripts/agent-telemetry.sh
+++ b/.claude/scripts/agent-telemetry.sh
@@ -587,7 +587,7 @@ END {
   # over-masking, which is the direction the governing asymmetry above demands.
   for (i = 1; i <= n; i++) {
     s = L[i]; out = ""
-    while (match(s, /-----BEGIN [^-]*PRIVATE KEY[^-]*-----/)) {
+    while (match(s, /-----BEGIN ([A-Z0-9][A-Z0-9. ]*)?PRIVATE KEY( BLOCK)?-----/)) {
       bs = RSTART; bl = RLENGTH
       head = substr(s, 1, bs - 1)
       # `scan`, deliberately NOT `tail`: pass B owns a global of that name. It
@@ -596,7 +596,7 @@ END {
       # U[] flag below, and this program has paid for that class enough times.
       scan = substr(s, bs + bl)
       depth = 1; closed = 0
-      while (depth > 0 && match(scan, /-----(BEGIN|END) [^-]*PRIVATE KEY[^-]*-----/)) {
+      while (depth > 0 && match(scan, /-----(BEGIN|END) ([A-Z0-9][A-Z0-9. ]*)?PRIVATE KEY( BLOCK)?-----/)) {
         mark = substr(scan, RSTART, RLENGTH)
         scan = substr(scan, RSTART + RLENGTH)
         if (mark ~ /^-----BEGIN/) depth++
@@ -685,7 +685,7 @@ END {
     close_at = 0; close_end = 0; bdepth = U[i]
     for (j = i + 1; j <= n && close_at == 0; j++) {
       bscan = L[j]; bdrop = 0
-      while (match(bscan, /-----END [^-]*PRIVATE KEY[^-]*-----/)) {
+      while (match(bscan, /-----END ([A-Z0-9][A-Z0-9. ]*)?PRIVATE KEY( BLOCK)?-----/)) {
         # Absolute end offset of this marker in L[j], accumulated as `bscan` is
         # consumed. The closing line is masked UP TO the marker that actually
         # balanced the depth, not the first one on the line — closing at the
@@ -793,9 +793,9 @@ redact() {
     -e 's/(gh[pousr]_[A-Za-z0-9]{4})[A-Za-z0-9]+/\1…<redacted>/g' \
     -e 's/(AKIA[0-9A-Z]{4})[0-9A-Z]+/\1…<redacted>/g' \
     -e 's/(xox[baprs]-[A-Za-z0-9]{4})[A-Za-z0-9-]+/\1…<redacted>/g' \
-    -e 's/-----BEGIN [^-]*PRIVATE KEY[^-]*-----([^-]|-[^-])*(-----END [^-]*PRIVATE KEY[^-]*-----)?/<redacted-private-key>/g' \
-    -e 's/-----(BEGIN|END) [^-]*PRIVATE KEY[^-]*-----/<redacted-private-key>/g' \
-    -e 's/(-----BEGIN [^-]*PRIVATE KEY[^-]*-----)[^-]*/\1<redacted-key-material>/g' \
+    -e 's/-----BEGIN ([A-Z0-9][A-Z0-9. ]*)?PRIVATE KEY( BLOCK)?-----([^-]|-[^-])*(-----END ([A-Z0-9][A-Z0-9. ]*)?PRIVATE KEY( BLOCK)?-----)?/<redacted-private-key>/g' \
+    -e 's/-----(BEGIN|END) ([A-Z0-9][A-Z0-9. ]*)?PRIVATE KEY( BLOCK)?-----/<redacted-private-key>/g' \
+    -e 's/(-----BEGIN ([A-Z0-9][A-Z0-9. ]*)?PRIVATE KEY( BLOCK)?-----)[^-]*/\1<redacted-key-material>/g' \
     -e 's/(eyJ[A-Za-z0-9_-]{6})[A-Za-z0-9_.-]{20,}/\1…<redacted-jwt>/g' \
     -e 's/((secret|token|password|passwd|api[_-]?key)["'"'"']?[[:space:]]*[:=][[:space:]]*["'"'"']?)[^"'"'"'[:space:],}]{8,}/\1<redacted>/gI'
 }
@@ -824,7 +824,7 @@ sha256_digest() {
 # form), each time reporting a real leak as "clean". The test suite asserts a
 # sample of EVERY numbered shape is both detected AND redacted; add to both
 # lists together or the test fails.
-CRED_RE='(github_pat_[A-Za-z0-9_]{20,}|gh[pousr]_[A-Za-z0-9]{16,}|AKIA[0-9A-Z]{12,}|-----BEGIN [^-]*PRIVATE KEY[^-]*-----|xox[baprs]-[A-Za-z0-9-]{10,}|eyJ[A-Za-z0-9_-]{10,}\.[A-Za-z0-9_-]{10,}|(secret|token|password|passwd|api[_-]?key)["'"'"']?[[:space:]]*[:=][[:space:]]*["'"'"']?[^"'"'"'[:space:],}]{8,})'
+CRED_RE='(github_pat_[A-Za-z0-9_]{20,}|gh[pousr]_[A-Za-z0-9]{16,}|AKIA[0-9A-Z]{12,}|-----BEGIN ([A-Z0-9][A-Z0-9. ]*)?PRIVATE KEY( BLOCK)?-----|xox[baprs]-[A-Za-z0-9-]{10,}|eyJ[A-Za-z0-9_-]{10,}\.[A-Za-z0-9_-]{10,}|(secret|token|password|passwd|api[_-]?key)["'"'"']?[[:space:]]*[:=][[:space:]]*["'"'"']?[^"'"'"'[:space:],}]{8,})'
 
 # TABLE variant of CRED_RE — identical shapes, but the prefix-identified ones
 # (gh?_ / github_pat_ / AKIA / xox / eyJ) are BOUNDARY-ANCHORED: the char before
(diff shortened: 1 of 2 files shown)


devantler:

## Validation record (detail deliberately kept out of the PR body)

**RED → GREEN, against the shipped `awk` program extracted from the script — not a copy.**

| Input | before | after |
|---|---|---|
| `-----BEGIN PGP PRIVATE KEY BLOCK-----` + body | body emitted **verbatim** | `<redacted-key-material>` |
| `-----BEGIN X9.42 DH PRIVATE KEY-----` + body | body emitted **verbatim** | `<redacted-key-material>` |
| `-----BEGIN RSA PRIVATE KEY-----` + body (control) | masked | masked |
| `-----BEGIN CERTIFICATE-----` + body (negative control) | untouched | untouched |
| prose containing `PRIVATE KEY` (negative control) | untouched | untouched |

**Detector leg, checked separately** — the redactor and detector are the pair whose divergence reports "clean" over a live leak, so both were verified against `CRED_RE`: PGP armor and the digit/punctuation label are now **detected**; `CERTIFICATE` is still not.

**Ablation — new tests run against the UNFIXED script.** Three of the six shape-9 rows fail without the production change (`9a` body masking, `9b` body masking, and the structural row). The three that pass in both states are the two negative controls and the "record after it survives" companion — a negative control that changed state would not be controlling anything.

**Real-corpus effect.** One new `private-key-block` row appears on a 1-day corpus scan. Provenance checked before drawing any conclusion: it is another session's **prose quoting this issue's own text**, not key material — verified by locating the record and confirming the surrounding content, without echoing it. So the widened detector is firing on a label family it was previously blind to, on real data, and no leak is implied.

**CI.** All checks green at `ac848c3`. The first attempt showed one `macos-latest` failure in an unrelated assertion (`a real sleep command is still counted`, efficiency section — unreachable from this diff); re-running the identical commit passed. That flake is recorded with evidence on #2661.

**User evaluation.** Exercised the real behaviour rather than reasoning about it: ran the miner end-to-end against the live corpus on both the base and changed scripts and diffed the safety section, confirming the only difference is the newly-detected label family and that no other count moved.

devantler:

## Correction: the first version of this fix introduced a leak, and self-review caught it

Recording this in full, because the failure is more instructive than the fix.

The first commit widened the label to `[^-]*PRIVATE KEY[^-]*`, arguing in its own comment that hyphen is the one character a PEM label may not contain, so the expression **"physically cannot run past the five-dash boundary"**. That argument is wrong in a way that reads as rigorous, and it produced two defects — both reproduced against the shipped `awk` program, not reasoned about:

**1. Under-mask — a real key body leaked (the serious one).**

The same edit widened the **END** marker. END is a *stop condition*, so loosening it can only ever mask **less**. A prose line between a truncated key's `BEGIN` and its body closed the span early:

```
-----BEGIN RSA PRIVATE KEY-----      first version        main (before)
MIIBODY_001                          <redacted>           <redacted>
-----END ... PRIVATE KEY-----        <redacted>           <redacted>
MIIBODY_002_LEAKED                   MIIBODY_002_LEAKED   <redacted>
MIIBODY_003_LEAKED                   MIIBODY_003_LEAKED   <redacted>
```

`-----END ... PRIVATE KEY-----` already occurs in the live corpus. A change whose entire purpose was to stop a leak had introduced one, in the direction this file's governing asymmetry explicitly forbids.

**2. Runaway plus a false leak report.**

The expression never needed to *cross* a dash — only to find a dash-**free gap** between two dash runs. `-----BEGIN was seen; the PRIVATE KEY is elsewhere -----` matched as a marker; the unpaired opener then masked the stream to EOF, and the same widening in `CRED_RE` reported a private-key leak that did not exist, in the section that tells the reader to rotate the credential.

## What shipped instead

`([A-Z0-9][A-Z0-9. ]*)?PRIVATE KEY( BLOCK)?` — bounded to the two families the specs require and no further. The label must **start with an alphanumeric**, which is what rejects `... ` and lowercase prose, and the OpenPGP suffix is the **literal** ` BLOCK` rather than a trailing wildcard.

## The test gap this exposed, now closed

The original controls could not have caught either defect: a `CERTIFICATE` label and a dash-free prose line behave identically under *every* candidate expression, so neither constrained the widening in the direction it actually broke. Three regression controls were added, and the grammar is now pinned from **both** sides — verified by running the suite against three variants:

| variant | result |
|---|---|
| too narrow — `[A-Z ]*` (i.e. `main`) | **fails** 9a, 9b, structural |
| too wide — `[^-]*…[^-]*` (first version) | **fails** all three regression controls + structural |
| shipped grammar | all nine pass |

A one-sided test would have accepted the leak. The structural row now pins **both** rejected spellings, so the over-wide one cannot reappear silently either.

devantler:



@coderabbitai review

devantler:

CodeRabbit completed without a gate-satisfying artifact at this head: it reported `Review limit reached` (free OSS review limit, next review in ~23 min) and carried the `rate limited by coderabbit.ai` marker, so no review ran. Its `✅ Action performed — Review finished` acknowledgement is not evidence to the contrary; that reply is emitted either way. Advancing to the next lane rather than serialising this PR behind an account-level window.




@codex review

devantler:

## Correcting my own claim above

I wrote that the prose end-marker `-----END ... PRIVATE KEY-----` **"already occurs in the live corpus."** I took that from the review without checking it. Having now measured it, the claim is true but materially weaker than I made it sound, so I am correcting it rather than leaving it to stand.

Measured across the whole session store: 4 files contain the literal. **Three of the four are this very session** — including the transcript of the review subagent that raised the finding. Exactly **one** pre-existing occurrence predates this run (dated 2026-08-03), and it is a subagent transcript, not a key-bearing record.

So the honest version is: **one** pre-existing occurrence, in agent prose about PEM markers — not a recurring corpus pattern.

**This does not change the P0 or the fix.** The defect never depended on that particular string: the widened `END` class admitted *any* dash-free prose between a truncated key's `BEGIN` and its body, including ordinary lowercase text, and `END` is a stop condition where loosening can only ever mask less. The corpus hit was supporting colour, and I over-weighted it.

Worth recording separately, because it will recur: **this tool's own corpus is self-contaminating.** Investigating a marker pattern writes that marker into the transcripts the tool then scans, so a naive "it appears in the corpus" check counts the investigation as evidence of the thing being investigated. The script's own comments already warn about this for the refusal detector; it applies to the key-marker detector too. Any future corpus-frequency claim here has to exclude the measuring session before it means anything.

devantler:

Requesting Cursor Bugbot — the third and final review lane. CodeRabbit returned `rate limited by coderabbit.ai` at this head, and Codex returned a usage-limit notice (`cr:no-gate@361a0414`, `codex:no-gate@361a0414`).

devantler:
@cursor review

devantler:

Requesting a review at the new head `7f0dbb3d` (CI green: 12 success / 22 skipped / 0 failing).

A local review round at the previous head found an **under-mask regression this PR introduced versus `main`** — `-----BEGIN··PRIVATE·KEY-----` (repeated space, no label word) masked on the base commit and leaked after the widening, because the replaced `[A-Z ]*` held the space inside its character class while the optional-group form demands `PRIVATE` immediately after the single literal space. Fixed in `7f0dbb3d` with ` *`, plus regression control 3 and the narrowed spelling pinned as a third rejected form in the structural assertion.

@coderabbitai review

devantler:

**Re: the RFC 7468 label-grammar finding — partially accepted (option 2), refuted on scope for option 1, follow-up filed as #2662.** Resolved at `9a9b882c`.

**The factual claim is correct and I am not disputing it.** RFC 7468 `labelchar` (%x21-2C / %x2E-7E) admits punctuation this expression rejects, and the grammar permits an internal hyphen-minus. Verified against the shipped program: `FOO/BAR PRIVATE KEY` and `FOO-BAR PRIVATE KEY` are matched by neither the redactor nor the detector.

**Refuting option 1 (widen the class) on scope, with reproduced evidence rather than preference.** Widening the label class is the precise change that produced the two defects this PR already documents — and the second is the one that matters here: a loosened `END` marker is a **stop condition**, so widening it can only ever mask **less**. A prose `-----END ... PRIVATE KEY-----` closed a truncated key's span early and emitted the remaining body verbatim. Your own note that option 1 "needs special care for END handling" is the same observation, and it is why this is not a regex change at all: it requires pairing each closer to its opener's label, which is a structural change to the span walker. Smuggling that into a fix for the label gap would repeat the mistake that made this PR necessary.

**Accepting option 2, which is the cheap and honest half.** The shape-9 comment quoted the RFC production and then shipped a narrower class, which reads as an *incomplete implementation* rather than a *bounded* one. `9a9b882c` states the boundary directly: what is deliberately out of scope, what is covered, and why closing the gap is a span-walker change.

**Coverage measured, not asserted** — all eight registered private-key labels mask:

| label | result |
|---|---|
| `PRIVATE KEY` · `ENCRYPTED PRIVATE KEY` | mask |
| `RSA` · `DSA` · `EC` · `OPENSSH PRIVATE KEY` | mask |
| `PGP PRIVATE KEY BLOCK` · `X9.42 DH PRIVATE KEY` | mask |

No known producer emits a punctuation or hyphen label for private-key material, so this is a residual spec gap, not a live leak — tracked at **#2662** (Security, boarded, sub-issue of #2626) with the END-pairing design and the #2659 regression controls as acceptance criteria.

Also note this review arrived as a plain issue **comment** with no review object at the head — the third time that shape has been recorded here, and the reason both surfaces are swept before any promotion decision.

devantler:

Requesting a review at `9a9b882c` — CI green (12 success / 22 skipped / 0 failing). This head adds only the scope clarification resolving the RFC 7468 finding above; the code change is unchanged from `7f0dbb3d`.

Lane state for whoever collects this: **Codex and Cursor Bugbot are both BILLING-limited** (`reached your Codex usage limits`, `Bugbot couldn't run - usage limit reached`) and will not heal on a timer — only an account admin can lift either. CodeRabbit is the sole lane that recovers on its own, and it served at 11:31Z, so its per-PR window may still be locked. If it refuses and the other two are still capped, a local review round is authorised.

@coderabbitai review

devantler:

**Re: the corpus-provenance finding** ([comment](https://github.com/devantler-tech/monorepo/pull/2659#issuecomment-3487817551), CodeRabbit review `74b124ee`, at `9a9b882c`) — **valid, and fixed** in `2720440e`.

The parenthetical at `.claude/scripts/agent-telemetry.test.sh:4388` claimed the widened END-marker shape *"already occurs in the live corpus"*. That reads as evidence of a recurring production pattern, and the measurement does not support it.

**Measured independently before fixing** (live corpus = `~/.claude/projects/**/*.jsonl`, 2026-08-04):

| Provenance | Occurrences |
|---|---|
| Predates this investigation | **1** — a `user`-role record of **2026-07-07** |
| Written by this investigation + sessions since discussing it | **all others** (~70 across 16 sessions) |

One correction to the finding's wording: that single pre-existing record is `role=user`, not agent prose — in this corpus `user`-role records carry tool-result text. It is transcript text either way, not an emitted key, so the conclusion is unchanged.

**The self-contamination is live, not theoretical.** While measuring, the file count rose **17 → 20** as this very session wrote the marker into its own transcript. The corpus stores the sessions that scan it.

The comment now states that provenance and claims only what the evidence supports — the shape is **reachable in ordinary text** — instead of implying production repeatedly emits it. The surrounding argument is unaffected: the single 07-07 record is still sufficient to show reachability, which is what failure mode 2 rests on.

**Verification.** Comment-only change; `bash -n` clean, no new shellcheck findings (the two `SC2010`s at lines 994/997 are pre-existing and untouched).

⚠️ **Worth recording separately: this suite is nondeterministic in a local multi-session environment.** Running the **unmodified** head twice gave **290** then **294** failures, with 5 tests flipping one way and 9 the other — it scans the live corpus while ~12 concurrent sessions write to it. All 5 tests that differed between my tree and the control appear in the control's own second run, so the change is behaviour-neutral. CI is green on this PR, so this is a local-environment property, not a regression here; filing it separately rather than widening this PR.

devantler:

Requesting a review at `2720440e` — CI green (0 failing, `mergeStateStatus: CLEAN`). The new commit resolves the corpus-provenance finding from the `74b124ee` review; resolution record and measurement are in the comment above.



@coderabbitai review

devantler:

**Readiness — promoting.** Verified at `2720440e`.

**1. Programmatically tested.** `CI - Required Checks` = **success** (the only required check on this repo's rulesets). 0 review threads, 0 non-thread/comment-shaped findings, no conflict with base.

> `mergeStateStatus` reads `UNSTABLE`, caused solely by the **`CodeRabbit` commit status still PENDING** — a non-required context. Every other check is `success` (12) or `skipped` (22). Noting it rather than letting it read as green: that status has been unreliable in *both* directions here (it read `success` at an unreviewed head on platform#2949 yesterday), so the substantive review artifact is what I judged on, not the status.

**2. Reviewed.** CodeRabbit at `2720440e`: *"I reviewed commit `2720440e`. I found no blocking issue."* Ack verb **`Review triggered`** — the accepted shape, not the `finished` + quota-block refusal shape — and no rate-limit marker in the summary.

**3. Tried and evaluated as a user.** This is a comment change, so the only meaningful user test is *does the documentation now tell the truth* — I re-measured it independently rather than taking the reviewer's word:

- Swept the live corpus myself: **exactly one** occurrence predates the investigation (`user`-role record, 2026-07-07); all others were written by the investigation and by sessions discussing it. That is what the comment now claims, and no more.
- Watched the self-contamination happen live — the matching-file count rose **17 → 20** during the measurement, as this session wrote the marker into its own transcript.
- Ran the suite **against a pristine control at the same base**, because a raw pass/fail here proves nothing: the suite is **nondeterministic in this environment**. The *unmodified* head gave **290** then **294** failures across two runs. All 5 tests that differed between my tree and the control appear in the control's own second run, so the change is behaviour-neutral — which is the correct expectation for a comment-only edit, and I would not have known it without the control.
- `bash -n` clean; no new shellcheck findings (the two `SC2010`s at lines 994/997 are pre-existing and untouched).

CI is green on this PR, so that local nondeterminism is an environment property (≈12 concurrent sessions writing the corpus it scans), not a regression here. Flagged in the resolution comment rather than widened into this PR.

PR #2675, opened 2026-08-05
fix(branch-cleanup): stop the pre-delete worktree re-check failing open

### LINKED ISSUE #2674: branch-cleanup's worktree keep-set re-check always fails open, so a live worktree's branch can be deleted

## Evidence

`branch-cleanup.sh` deletes local branches. Immediately before each delete it re-checks whether a worktree has claimed the branch since the keep-set snapshot — the last line of defence against deleting a branch that a live session is sitting on. That check cannot return true.

```sh
# .claude/scripts/branch-cleanup.sh:370
if printf '%s\n' "$wt_list" | grep -Fxq "branch refs/heads/$b"; then
```

The script runs under `set -uo pipefail` (line 36). `grep -Fxq` exits at the first matching line; the `printf` is still writing the rest of the worktree list, so it dies of `SIGPIPE` (141), and `pipefail` makes that the pipeline's status. A branch that **is** checked out is therefore reported as checked out **nowhere**, and execution falls through to `git update-ref -d`.

Measured directly, varying only the number of worktree entries (match placed early, as a real match usually is):

| worktree entries | payload | verdict |
|---|---|---|
| 0 | 113 B | KEEP (correct) |
| 1 | 296 B | KEEP (correct) |
| 2 | 479 B | KEEP (correct) |
| 5 | 1.0 KB | **DELETE (wrong)** |
| 100 | 18.6 KB | **DELETE (wrong)** |
| 2400 | 451 KB | **DELETE (wrong)** |

The threshold is about **1 KB — five worktree entries**. This host currently reports **372 lines** from `git worktree list --porcelain`, roughly 124 worktrees, so the check has been failing open here continuously, not marginally.

The n≤2 rows are the control: the predicate does work on a tiny list, which is why this survived review and why an interactive spot-check would confirm it as fine.

## Why it matters

The comment directly above the line states the intent — *"update-ref would then delete a branch live in another worktree, leaving that worktree on a dangling/unborn HEAD"* — and the preceding lines go out of their way to fail closed when the enumeration errors. That care is real; the very next line then discards it.

Scope, stated honestly: this is the **secondary** re-check. The primary keep-set is built earlier and consulted through `re_kept()`, which greps files directly and is **not** affected. So the exposure is a branch whose worktree appeared *after* the snapshot — the exact race this line exists to cover. It is a defence-in-depth layer that has never functioned, not an unguarded delete path.

## Fix

Remove the pipe so the status is grep's alone:

```sh
if grep -Fxq "branch refs/heads/$b" <<<"$wt_list"; then
```

Do **not** fix this by dropping `pipefail` — that would hide genuine failures in intermediate stages elsewhere in the script.

## Wider scope

The same construct is present at **81** places across 6 scripts under `pipefail` (`agent-telemetry.sh` 6, `board-add.test.sh` 16, `branch-cleanup.sh` 1, `review-provider-loop-contract.test.sh` 3, `worktree-cleanup-all.test.sh` 14, `worktree-cleanup.test.sh` 41). monorepo#2673 fixes the 132 in `agent-telemetry.test.sh` and adds a reusable lint for the shape; the rest are untouched. `worktree-claim.test.sh` also has 2 but sets no `pipefail`, so it is unaffected.

## Acceptance criteria

- The line-370 re-check returns true for a branch present in a realistically-sized worktree list, proven by a test that fails on the current code.
- The remaining 80 instances are triaged: fixed, or recorded as unreachable with the reason.
- The lint from monorepo#2673 is applied to these scripts so the shape cannot return.

Rough size: S for the safety fix, M including the sweep.

devantler:

## Why

`branch-cleanup.sh` deletes local branches. The last thing it does before each delete is re-read the live worktree list, to catch a branch a session claimed since the keep-set snapshot. That check could never return true, so the protection it provides has never existed.

The consequence is a branch deleted out from under a running session, leaving that worktree on a dangling HEAD — which is precisely the outcome the code comment above the line says it is preventing.

## What

The check ran under `pipefail` and piped into a quiet `grep`. `grep` stops at the first match while the writer is still going, the writer dies of `SIGPIPE`, and `pipefail` reports *that* as the answer — so "is checked out" came back as "is checked out nowhere". It now feeds grep directly.

It is size-dependent, which is why it read as correct: fine for a one- or two-entry worktree list, wrong from about five entries up. This host enumerates roughly 124.

Scope, stated plainly: this is the **secondary** re-check. The primary keep-set is built earlier and is unaffected, so the exposure was a worktree appearing *after* the snapshot — the exact race this line covers. A defence-in-depth layer that was fully non-functional, not an unguarded delete path.

The new test drives the real script with a `git` shim that returns a bigger worktree list at the re-check than at the snapshot. Before the fix it fails, with `line 370: printf: write error: Broken pipe` in the output; after it passes. A second arm deletes an *unclaimed* branch at the same list size and passes either way, so the first arm cannot pass vacuously.

Same root cause as #2673, different file and a live safety impact rather than a test-harness one, so it ships separately. 80 further instances of the shape remain across other scripts and are tracked on the issue.

Fixes #2674

### DIFF
--- a/.claude/scripts/branch-cleanup.sh
+++ b/.claude/scripts/branch-cleanup.sh
@@ -367,7 +367,15 @@ if [ "$DO_LOCAL" -eq 1 ]; then
         echo "$SLUG: keep '$b' — worktree enumeration failed; refusing to delete (fail-closed)" >&2
         l_keep=$((l_keep+1)); errors=$((errors+1)); continue
       fi
-      if printf '%s\n' "$wt_list" | grep -Fxq "branch refs/heads/$b"; then
+      # Feed grep directly instead of through a pipe. This script runs under
+      # `pipefail`, and `grep -Fxq` exits at the first match — so with a pipe the
+      # still-writing `printf` takes SIGPIPE (141) and pipefail reports THAT as
+      # the pipeline status, turning "is checked out" into "is checked out
+      # nowhere" and deleting the branch this very check exists to protect.
+      # Size-dependent, hence easy to miss: correct up to ~2 worktree entries,
+      # wrong from ~5 up. Do NOT "simplify" this back into a pipeline, and do
+      # not fix such a case by dropping pipefail (monorepo#2674).
+      if grep -Fxq "branch refs/heads/$b" <<<"$wt_list"; then
         echo "$SLUG: keep '$b' — checked out by a worktree since the snapshot" >&2
         l_keep=$((l_keep+1)); continue
       fi
--- a/.claude/scripts/branch-cleanup.test.sh
+++ b/.claude/scripts/branch-cleanup.test.sh
@@ -411,6 +411,74 @@ report "apply recorded the remote deletion sha and MERGED status" \
 # Silence unused-var lint for the open_sha we only needed for push tip creation.
 : "$open_sha" "$local_sha"
 
+# --- 13. the pre-delete worktree re-check must survive a realistic list -----
+# monorepo#2674. The local delete loop re-reads `git worktree list --porcelain`
+# immediately before `update-ref -d`, to catch a worktree that claimed the
+# branch AFTER the keep-set snapshot. That predicate was a pipeline ending in
+# `grep -Fxq`, and this script runs under `pipefail`: grep exits at the matching
+# line, the still-writing `printf` takes SIGPIPE (141), and pipefail reports 141
+# as the pipeline status — so a branch that IS checked out reads as checked out
+# NOWHERE and gets deleted, leaving that worktree on a dangling HEAD.
+#
+# It is a SIZE-dependent failure, which is why it survived review: the predicate
+# is correct for a one- or two-entry list and wrong from about five entries up.
+# A `git` shim on PATH returns the real enumeration for the keep-set snapshot
+# and a larger, later one for the re-check — exactly the race the re-check
+# exists to cover, and the only way to isolate it from the primary keep-set.
+wt_shim_dir="$tmp/wtshim"; mkdir -p "$wt_shim_dir"
+real_git=$(command -v git)
+cat >"$tmp/bin/git" <<SHIM
+#!/usr/bin/env bash
+if [ "\$1" = "worktree" ] && [ "\$2" = "list" ] && [ -n "\${WT_SHIM_STATE:-}" ]; then
+  n=\$(cat "\$WT_SHIM_STATE" 2>/dev/null || echo 0); echo \$((n+1)) >"\$WT_SHIM_STATE"
+  if [ "\$n" -ge 1 ] && [ -s "\${WT_SHIM_INJECT:-/nonexistent}" ]; then
+    cat "\$WT_SHIM_INJECT"; exit 0
+  fi
+fi
+exec "$real_git" "\$@"
+SHIM
+chmod +x "$tmp/bin/git"
+
+# A worktree list of realistic size. The target branch sits near the FRONT,
+# as a real match usually does — that is what leaves the writer with work to do.
+mk_wt_list() {   # $1 = branch to include ("" for none)
+  local target="$1" i
+  printf 'worktree %s\nHEAD %s\nbranch refs/heads/main\n\n' "$work" "$(git -C "$work" rev-parse main)"
+  [ -n "$target" ] && printf 'worktree %s/live\nHEAD %s\nbranch refs/heads/%s\n\n' "$tmp" "$(git -C "$work" rev-parse main)" "$target"
+  for i in $(seq 1 40); do
+    printf 'worktree %s/.claude/worktrees/sess-%s\nHEAD %s\nbranch refs/heads/claude/other-session-%s\n\n' \
+      "$work" "$i" "$(git -C "$work" rev-parse main)" "$i"
+  done
+}
+
+# ARM 1 — the guard: branch IS in the later list, so it must be KEPT.
+wt_sha=$(mk_remote_branch "claude/wt-reclaim-2674")
+git -C "$work" checkout main >/dev/null
+: >"$OPEN_HEADS_FILE"
+printf '%s\tMERGED\t%s\n' "claude/wt-reclaim-2674" "$wt_sha" >"$PR_EVIDENCE_FILE"
+export WT_SHIM_STATE="$wt_shim_dir/count" WT_SHIM_INJECT="$wt_shim_dir/list"
+: >"$WT_SHIM_STATE"; mk_wt_list "claude/wt-reclaim-2674" >"$WT_SHIM_INJECT"
+manifest="$tmp/manifest-2674a"; : >"$manifest"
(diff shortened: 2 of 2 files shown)


devantler:


@coderabbitai review

devantler:



CodeRabbit did not acknowledge the request at this head (no reaction, no comment, status still the
push-time skipped marker) — the account is rate limited. Codex is billing-limited. Advancing to
Cursor Bugbot, which is the next lane in priority order. This is a live safety fix, so it is worth
the monthly quota.

devantler:
@cursor review

devantler:

## Readiness evidence — exercised, not reasoned about

**Head:** `509120e4ccd90601fd672a72725979360e164665`

**1. Programmatically tested — RED then GREEN, on the real script.**

| | result |
|---|---|
| `main`'s `branch-cleanup.sh` + this test | `FAIL: worktree re-check keeps a branch claimed after the snapshot` — with `line 370: printf: write error: Broken pipe` in the captured output; suite exits 1 |
| this branch | both assertions pass; suite exits 0; zero broken-pipe warnings |

`Test branch cleanup` is green on `ubuntu-latest` and `macos-latest`, so the fix is not host-specific.

**2. The control is the point.** A test asserting only "the branch survived" would pass on a harness that never deletes anything. The second arm deletes an **unclaimed** branch at the same worktree-list size and passes in **both** arms, so arm 1 cannot pass vacuously.

**3. Used as its user.** The user here is a scheduled agent ending a tick. I measured the predicate directly against realistic input before touching the code, varying only the number of worktree entries:

| entries | payload | verdict |
|---|---|---|
| 0 / 1 / 2 | ≤479 B | KEEP (correct) |
| 5 | 1.0 KB | **DELETE (wrong)** |
| 100 | 18.6 KB | **DELETE (wrong)** |
| 2400 | 451 KB | **DELETE (wrong)** |

This host reports **372 lines** from `git worktree list --porcelain`, so the check has been failing open here continuously. The n≤2 rows are why it read as correct on inspection.

**4. Acted on the finding beyond the diff.** Because the fix was not yet merged, I **deliberately skipped this tick's `branch-cleanup.sh` sweep** rather than run a script I had just proven can delete a branch a live session is on. Recorded in the run notes; the sweep resumes once this merges.

**5. Self-review caught one defect before review.** The test's `git` shim read a truncated counter file, making `[ "" -ge 1 ]` print `integer expression expected` to stderr — captured into the assertion's `$out`, where it would sit beside genuine failure text. Fixed in `509120e4`; RED/GREEN re-verified afterwards.

devantler:


@coderabbitai review

devantler:


@coderabbitai review

devantler:


@coderabbitai review

devantler:


@coderabbitai review

devantler:

## Independent verification at `509120e4` (RED/GREEN reproduced, and the class swept)

**RED/GREEN, run here rather than inherited.** Took the new test against the *unfixed* script on a
clean detached worktree at `origin/main`:

- old script + new test → `FAIL: worktree re-check keeps a branch claimed after the snapshot (#2674)`,
  carrying the literal `branch-cleanup.sh: line 370: printf: write error: Broken pipe` — the exact
  SIGPIPE mechanism the PR describes, not a proxy for it
- the **control arm still PASSED** in that same run, so the harness is not vacuously failing everything
- fixed script + new test → both arms pass; full suite `all assertions passed`

**Swept the defect class rather than the single site.** Checked whether the same
pipe-into-quiet-consumer shape survives elsewhere in the script:

- `re_kept()` (line 394) greps **files**, not a pipe — no SIGPIPE exposure (it only looks like a hit
  because of the `||`)
- on line 350, `git branch -r --contains … | head -1` sits inside a command substitution whose exit
  status is **discarded** — `[ -z "$(…)" ]` tests the captured string, and the captured first line is
  what is wanted whether or not the writer is signalled

So this is the only site of the class in the file, and the fix is complete rather than partial.

**Also confirmed** `#!/usr/bin/env bash` + `pipefail` in both touched files, so the `<<<` here-string
is valid — under a `#!/bin/sh` shebang it would have been a syntax error rather than a fix.

Pentad at this head: CI green, 0 unresolved threads, 0 non-thread findings, `CLEAN`. The only missing
condition is a green review — Codex and Bugbot are both hard usage/billing limited today, and
CodeRabbit is serving roughly one review per ~10-minute window.

devantler:


@coderabbitai review

devantler (commented):

**Reviewed commit:** `509120e4ccd90601fd672a72725979360e164665`

### Per-lane evidence (direct, per-PR — not inherited from a portfolio digest)

| Lane | State at this head | Evidence |
|---|---|---|
| CodeRabbit | rate limited, 4× on this exact commit | statuses `11:38:16Z`, `11:44:19Z`, `11:46:42Z` (+ a concurrent instance's at `11:34:10Z`), each `state=success` with description `Review rate limited`. It served exactly one review this hour (on #2673) and refused every subsequent request. |
| Codex | usage/billing limit, no retry window | `You have reached your Codex usage limits for code reviews` — the account needs credits enabled; only the maintainer can lift it. |
| Cursor Bugbot | usage/billing limit, no retry window | `Bugbot couldn't run - usage limit reached` posted on this PR at `10:45:05Z`; an admin must raise it in the Cursor dashboard. |

CodeRabbit is serving roughly one review per ~10-minute window and is shared with a concurrent
instance, so waiting would park this finished change past the end of this run. Reviewing locally per
the 2026-07-21 direction rather than leaving a verified safety fix blocked on a provider quota.

### What I actually checked

1. **RED/GREEN reproduced here, not inherited.** New test against the *unfixed* script on a clean
   detached worktree at `origin/main` → arm 1 FAILS carrying the literal
   `printf: write error: Broken pipe`; the **control arm still passes** in that same run, so the
   harness is not vacuously failing. Fixed script → both arms pass, full suite green.
2. **Correctness of the substitution.** `printf '%s\n' "$x" | grep -Fxq` and `grep -Fxq <<<"$x"`
   both present the value with exactly one trailing newline, so the match semantics are unchanged;
   only the SIGPIPE exposure is removed.
3. **Shell validity.** Both touched files are `#!/usr/bin/env bash` with `pipefail` set — under a
   `#!/bin/sh` shebang the `<<<` here-string would be a syntax error rather than a fix.
4. **Defect class swept, not just the reported site.** `re_kept()` greps files (no pipe); line 350's
   `| head -1` is inside a command substitution whose exit status is discarded, so its captured first
   line is correct regardless. This is the only site of the class in the file.
5. **Safety framing.** The guarded predicate is the pre-delete worktree re-check, so the pre-fix
   behaviour deletes a branch that IS checked out — matching the repo rule that the keep-set is the
   whole safety property of this script.

### Not blocking, recorded for the record

The test shims `git` on `PATH` and relies on a call-count file to distinguish the snapshot read from
the re-check read. That is the only way to isolate this race, and the shim is removed at the end of
the block, but it does make the assertion order-sensitive — worth knowing if a future test is
inserted between the two arms.

PR #2695, opened 2026-08-06
fix(scripts): refuse repair when a submodule path resolves to the superproject gitdir

### LINKED ISSUE #2694: End-of-tick branch hygiene fail-closes on every tick: a submodule init redirects the monorepo main checkout via config.worktree

## Evidence

`branch-cleanup.sh` refused the end-of-tick sweep on 2026-08-06 (improver tick 990):

```
monorepo: ABORT — .../monorepo is not the repository root; git resolved upward to a
  parent repository, so this sweep would delete from the wrong tree.
```

The abort is correct and the guard is working. The cause is that the monorepo main checkout resolves its working tree somewhere else entirely:

```
git -C .../monorepo rev-parse --show-toplevel
  -> .../monorepo/.claude/worktrees/<session-slug>/platform
```

Traced to `core.worktree`:

```
git -C .../monorepo config --show-origin --get core.worktree
  file:.git/config.worktree   .../monorepo/.claude/worktrees/<session-slug>/platform
```

`extensions.worktreeConfig` is `true`, so `.git/config.worktree` is the **sanctioned** location for this key — the defect is not where it lives but that its **value redirects the main checkout into a platform submodule worktree**. Timestamps place it inside a live session: the worktree directory was created at 12:15 local and `.git/config` was rewritten at 12:18, with a `claude` process still holding that directory as its cwd. This is the `git submodule update --init` regression already described in the worktree-isolation notes, landing on the monorepo main worktree rather than on a submodule.

## Impact

Any git operation targeting the monorepo main checkout resolves to the wrong tree. The visible consequence is that **contract-mandated end-of-tick branch hygiene cannot run at all** — every tick fail-closes and reports nothing was swept, so `claude/*` and `cursor/*` branches accumulate unbounded. It fails safely rather than deleting from the wrong tree, which is why it has gone unnoticed: the sweep is skipped silently from the callers point of view.

## Expected behaviour

The monorepo main checkout resolves to its own root, and `branch-cleanup.sh` runs. A submodule init must not write `core.worktree` into the *parent* repositorys per-worktree config.

## Acceptance criteria

- `git -C <monorepo> rev-parse --show-toplevel` returns the monorepo root.
- `branch-cleanup.sh <monorepo> monorepo <manifest> dry-run claude` completes instead of aborting.
- The path that writes this key is identified and prevented from targeting the parent repo — `submodule-init.sh` is the fail-closed entry point and should cover this direction too, not only the submodule direction.
- A regression check that would fail if a parent repo again resolves outside its own root.

## Deliberately not fixed in this tick

Unsetting the key is a one-line reversible repair, but a sibling session was **actively running in that worktree** at the time (process holding it as cwd, config written minutes earlier). Changing shared git state under a live sibling is the cross-instance breakage the contract forbids, so this is prepared and handed over rather than applied mid-flight. Apply it between runs.

Rough size: S/M.

devantler:

## Why

The helper every agent run is told to use instead of a bare `git submodule update --init` could
silently break the **parent** repository. Handed a registered submodule directory that is non-empty
but not actually initialised, it wrote the wrong setting into the monorepo's own git config and
pointed the monorepo's main checkout at that submodule directory — then exited reporting success.

This happened for real on the agent host this morning: the monorepo checkout resolved to a submodule
folder, `git status` showed 182 files as deleted that were not, and the end-of-tick branch cleanup
every run is required to perform had been failing closed on **every** tick as a result. Because the
helper reported success, nothing surfaced it.

## What

The helper now checks the git directory it actually resolved and refuses when that turns out to be
the parent repository's, instead of writing into it. The refusal says what to do next. A correctly
initialised submodule repairs exactly as before.

The live symptom on the host was repaired separately; this stops it recurring.

Fixes #2694

### DIFF
--- a/.claude/scripts/submodule-init.sh
+++ b/.claude/scripts/submodule-init.sh
@@ -147,8 +147,14 @@ repair() {
   # place by a different route, so test the RESOLVED gitdir rather than the presence of a `.git` entry
   # (an existence check is subsumed by this one and provably never fires on its own). Fail closed: the
   # blast radius is the parent repo and every session sharing it (monorepo#2694).
-  super_mdir=$(git rev-parse --path-format=absolute --git-common-dir 2>/dev/null || true)
-  if [ -n "$super_mdir" ] && same_dir "$mdir" "$super_mdir"; then
+  # No `|| true` and no silent fallback: `same_dir` fails closed on an empty path, which here would
+  # SKIP the guard rather than trip it, so a failed probe must stop the run instead of quietly
+  # disabling the check. `errexit` covers a non-zero rev-parse; the emptiness test covers a
+  # zero-exit-no-output result.
+  super_mdir=$(git rev-parse --path-format=absolute --git-common-dir)
+  [ -n "$super_mdir" ] ||
+    die "cannot resolve the superproject's gitdir — refusing to repair '$path' rather than skip the parent-escape check"
+  if same_dir "$mdir" "$super_mdir"; then
     die "'$path' resolves to the SUPERPROJECT's gitdir ('$mdir'), not its own — refusing to repair, because that would redirect the parent repository's checkout at '$path'. Populate the submodule first: git submodule update --init '$path'"
   fi
 
--- a/.claude/scripts/submodule-init.sh
+++ b/.claude/scripts/submodule-init.sh
@@ -155,7 +155,7 @@ repair() {
   [ -n "$super_mdir" ] ||
     die "cannot resolve the superproject's gitdir — refusing to repair '$path' rather than skip the parent-escape check"
   if same_dir "$mdir" "$super_mdir"; then
-    die "'$path' resolves to the SUPERPROJECT's gitdir ('$mdir'), not its own — refusing to repair, because that would redirect the parent repository's checkout at '$path'. Populate the submodule first: git submodule update --init '$path'"
+    die "'$path' resolves to the SUPERPROJECT's gitdir ('$mdir'), not its own — refusing to repair, because that would redirect the parent repository's checkout at '$path'. The directory has content but no usable '.git', so nothing here is a real submodule checkout: remove its stray contents, then re-run 'submodule-init.sh $path' to populate it at the pinned commit"
   fi
 
   git config -f "$mdir/config" extensions.worktreeConfig true
--- a/.claude/scripts/submodule-init.sh
+++ b/.claude/scripts/submodule-init.sh
@@ -132,11 +132,25 @@ check_existing_worktrees() {
 # Move the stray shared core.worktree into each worktree's own per-worktree config. Idempotent.
 repair() {
   local path=$1
-  local mdir tree
+  local mdir tree super_mdir
+
   mdir=$(module_dir "$path")
   tree=$(module_tree "$path")
   { [ -n "$mdir" ] && [ -d "$mdir" ] && [ -n "$tree" ]; } ||
     die "git does not report a gitdir for '$path' (not an initialised submodule?)"
+
+  # `git -C <dir>` walks UP when <dir> is not itself a repository, so a REGISTERED submodule path that
+  # is non-empty but NOT initialised resolves to the SUPERPROJECT's gitdir — and repair would then pin
+  # `core.worktree` in the PARENT repository's per-worktree config, redirecting the parent's own main
+  # checkout at this directory. `is_populated` cannot catch it: it only asks whether the directory has
+  # entries, and one leftover file is enough. A `.git` that exists but points outward reaches the same
+  # place by a different route, so test the RESOLVED gitdir rather than the presence of a `.git` entry
+  # (an existence check is subsumed by this one and provably never fires on its own). Fail closed: the
+  # blast radius is the parent repo and every session sharing it (monorepo#2694).
+  super_mdir=$(git rev-parse --path-format=absolute --git-common-dir 2>/dev/null || true)
+  if [ -n "$super_mdir" ] && same_dir "$mdir" "$super_mdir"; then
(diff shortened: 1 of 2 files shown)


devantler:


@coderabbitai review

devantler:

CodeRabbit is rate limited at this head (status description `Review rate limited`, zero review objects), so the lane cannot deliver. Advancing to Codex.



@codex review

devantler:

Codex reports usage limits exhausted for code reviews at this head. Advancing to Cursor Bugbot.

devantler:
@cursor review

devantler:


@coderabbitai review

devantler:

CodeRabbit rate limited at this head too.



@codex review

devantler:

devantler:
@cursor review

.claude/scripts/submodule-init.sh
```diff
@@ -132,12 +132,32 @@ check_existing_worktrees() {
...
+  # checkout at this directory. `is_populated` cannot catch it: it only asks whether the directory has
+  # entries, and one leftover file is enough. A `.git` that exists but points outward reaches the same
+  # place by a different route, so test the RESOLVED gitdir rather than the presence of a `.git` entry
+  # (an existence check is subsumed by this one and provably never fires on its own). Fail closed: the
+  # blast radius is the parent repo and every session sharing it (monorepo#2694).
+  # No `|| true` and no silent fallback: `same_dir` fails closed on an empty path, which here would
+  # SKIP the guard rather than trip it, so a failed probe must stop the run instead of quietly
+  # disabling the check. `errexit` covers a non-zero rev-parse; the emptiness test covers a
+  # zero-exit-no-output result.
+  super_mdir=$(git rev-parse --path-format=absolute --git-common-dir)
```
devantler, line 154: nit — this reads the superproject gitdir from the **process cwd**, so it is correct only because every caller reaches `repair()` after the top-level `cd "$super_root"`. That coupling is implicit rather than enforced by the signature.

Not raising it above nit: `repair()` has exactly one caller (`init_repair_probe`), which is dispatched after that `cd`, and the suite exercises the guard from a main checkout **and** from a linked superproject worktree, where the two gitdirs differ in the way that matters. Threading `super_root` through the signature would make it explicit at the cost of a wider change than this fix warrants.

devantler (commented):

**Reviewed commit:** `3080d4487b1e92a78a4835c1fd1d785114238377`

Per-lane evidence, all taken directly on this PR at this exact head:

- **CodeRabbit** — commit status `success` with description `Review rate limited`; zero review objects at head. Requested 11:34:31Z, refused. (Also refused at the previous head 11:29Z.)
- **Codex** — `chatgpt-codex-connector[bot]` replied that the account has reached its usage limits for code reviews. Requested 11:35:52Z. No review object, no reviewed-commit marker.
- **Cursor Bugbot** — check-run `Cursor Bugbot` `conclusion=neutral` with `output.title=Error` (the run-never-ran shape, not findings), paired `cursor[bot]` comment "Bugbot couldn't run - usage limit reached", zero inline comments. Requested 11:35:52Z with a bare trigger.

### What I checked

Correctness of the new guard, its failure direction, and whether it can fire on a legitimate repair. A properly initialised submodule's gitdir lives under `<super>/.git/modules/...` (or `<super>/.git/worktrees/<wt>/modules/...` from a linked worktree), which is a different directory from the superproject's own common dir — so the identity test cannot refuse a valid repair, and the suite exercises both invocation contexts.

The guard is proven load-bearing rather than assumed: ablating it alone (2 lines changed, verified) turns 6 assertions red across both reproduction routes. An earlier `.git`-existence guard was dropped after ablation showed it never fires on its own.

Security direction is a tightening only — it removes a write into a repository the script was never meant to modify, and adds no new write, input, or privilege.

**Verdict: no P0/P1 findings** (1 nit, inline).

.claude/scripts/submodule-init.sh
devantler: > 🤖 Generated by the Agentic Engineer

Refuted as a change, recorded as the reason. `repair()` has exactly one caller, `init_repair_probe`, and every dispatch path reaches it after the top-level `cd "$super_root"`, so the cwd is not ambient — it is established two lines above the dispatch. The suite pins the property that actually matters: the guard is exercised from a main checkout (cases 12, 12c) and from a linked superproject worktree (case 5), where the submodule gitdir lives under `.git/worktrees/<wt>/modules/sub` and the comparison must NOT fire. Ablating the guard turns both reproduction routes red, and case 5 stays green with it in place.

Threading `super_root` through the signature would widen a one-concern fix without changing any reachable outcome.

devantler:

## Readiness at `3080d448`

**Programmatically tested.** Both reproduction routes verified RED before the guard and RED again when the guard alone is ablated (2 lines changed, count asserted — my first ablation attempt silently matched nothing and every assertion 'passed', which is why the changed-line count is now checked). Full pre-existing suite unaffected; shellcheck clean; `test-submodule-init` verified as actually RUN (macos + ubuntu, both success) rather than skipped, on every head.

**Reviewed.** Local review round 4874178839, clean at exactly this head, after all three lanes refused here with direct per-PR evidence. Its one nit was refuted with reasoning and the thread resolved.

**Tried and evaluated as a user.** Ran the fixed script against the *actual* host directory that caused the outage — the exact invocation that corrupted the monorepo now refuses with an actionable message, and the monorepo's `core.worktree` and toplevel were confirmed unchanged across the run. Then followed the remedy the message prints, end to end, on a throwaway fixture: refusal → clear stray contents → re-run → `isolated ✓`. That second step is what caught the message originally naming a command this repo's own contract says to use the helper instead of.

Self-review also caught a fail-open in my own first draft: `|| true` on the guard's probe made a failed `rev-parse` **skip** the guard rather than trip it.

Issue #2522, opened 2026-07-27
Credential table cannot separate a blob chance-match from a real credential without base64-run evidence


devantler:
> 🤖 Generated by the Agentic Engineer

## Evidence

#2471 asked for matches inside a base64 run to be separated from delimiter-anchored ones. An attempt in #2520 keyed that on the boundary character — `+` or `/` — reasoning that every other base64 character is in `[A-Za-z0-9_-]`, which the boundary anchor rejects, so those two are the only ones that can match mid-blob.

That reasoning is sound in one direction and was applied in the wrong one. Only `+`/`/` **can** match mid-blob; that does not make `/` **mean** mid-blob. Reproduced counterexample:

```
GET https://example.test/session/eyJhbGciOiJIUzI1NiJ9.eyJzdWIiOiIxMjM0NTY3ODkwIn0 HTTP/1.1
```

The JWT takes `/` as its boundary character, so the label would have marked a **genuinely exposed credential** as "very likely a chance substring of an encoded blob" — lowering triage priority on exactly the case the detector exists to catch. The label was removed from #2520 rather than shipped.

## Impact

The separation #2471 asked for is still missing, so blob chance-matches continue to sit in the same rows as real credentials. Measured 2026-07-25: 35 of 113 GitHub-token-shaped matches were chance substrings inside Codex `encrypted_content` blobs.

## Expected behaviour

A match is treated as blob-embedded only on **evidence of a surrounding base64 run**, never on the boundary character alone — and a real credential in ordinary URL or path text is never downgraded.

## Verified starting point

A run requirement discriminates both measured cases:

```
[A-Za-z0-9+/=]{16,}[+/]
```

- base64 blob containing a token shape → matches
- JWT as a URL path segment → does not match (`https://example.test/session` has `.`/`:` breaking the run, and `session` is 7 chars)

The obstacle is plumbing, not detection: the table extracts with `grep -o`, which discards the surrounding context, so the run evidence needs a separate per-file pass whose results are intersected with the extracted values.

## Also in scope (same divergence class)

The concentration line and `--credential-provenance` locator added in #2520 scan raw file lines, while the table structurally excludes complete `input_image.image_url` base64 payloads — encoded binary that manufactures token shapes at random `+`/`/` boundaries. So a record inside an excluded image can appear in the locator while the table stays empty. #2520 documents this divergence in both directions; deriving all three surfaces from the same filtered input would remove it.

## Acceptance criteria

[...]

PR #2707, opened 2026-08-06
fix(telemetry): require base64-run evidence before labelling a match blob-embedded

devantler:

## Why

The leak table is how we notice a credential reaching a transcript. Today it can't separate a real one from noise: a token-shaped string that occurs *by chance* inside an encoded blob lands in the same rows as a genuinely exposed credential. Measured last month, 35 of 113 GitHub-token-shaped matches were chance substrings of encoded data — so roughly a third of the table is noise nobody can tell apart from the thing it exists to catch.

An earlier attempt at this was closed rather than shipped, because it would have made the detector *worse* in the one direction that matters: it could have marked a genuinely exposed credential as "probably just encoding noise", quietly lowering triage priority on a real leak.

## What

Separates the two, but only on positive evidence that the match really sits inside encoded data — never on a guess from a single character, which is what sank the earlier attempt. Anything ambiguous stays in the plain high-signal row.

Nothing is suppressed or filtered out: a labelled row keeps its shape and its count, so the numbers are unchanged and only the triage order improves. The scan cost is unchanged too.

Part of #2522 — this delivers the classification half. The issue's remaining item (having the locator and concentration lines derive from the same filtered input as the table, so the three surfaces cannot contradict each other) is untouched and keeps the issue open.

## Blocked — GitHub Actions outage

GitHub reported **Actions: `major_outage`** (verified live at 2026-08-06T22:03Z), so CI cannot run at this head. Validated locally instead: the full contract suite is **520 passed / 0 failed** (baseline 515 + 5 new), with the new behaviour proven RED before GREEN and two ablations each firing their own named assertion.

### DIFF
--- a/.claude/scripts/agent-telemetry.sh
+++ b/.claude/scripts/agent-telemetry.sh
@@ -247,13 +247,20 @@ PROVTMP=$(mktemp "${TMPDIR:-/tmp}/.agtel_prov.XXXXXXXX") || { echo "cannot creat
 CONCTMP=$(mktemp "${TMPDIR:-/tmp}/.agtel_conc.XXXXXXXX") || { echo "cannot create temp file" >&2; exit 3; }
 CREDCONC=$(mktemp "${TMPDIR:-/tmp}/.agtel_credconc.XXXXXXXX") || { echo "cannot create temp file" >&2; exit 3; }
 CREDPROV=$(mktemp "${TMPDIR:-/tmp}/.agtel_credprov.XXXXXXXX") || { echo "cannot create temp file" >&2; exit 3; }
+# Blob-embedded evidence set (#2522): the subset of table values that were found
+# inside a base64 run. Holds normalised credential values, so it is created with
+# mktemp's private mode and removed by the same traps as every other scratch.
+CREDBLOB=$(mktemp "${TMPDIR:-/tmp}/.agtel_credblob.XXXXXXXX") || { echo "cannot create temp file" >&2; exit 3; }
+# Raw credential-table matches from the SINGLE decode pass, partitioned after
+# the fact into the plain and blob-embedded sets (#2522).
+CREDMATCH=$(mktemp "${TMPDIR:-/tmp}/.agtel_credmatch.XXXXXXXX") || { echo "cannot create temp file" >&2; exit 3; }
 # Its OWN scratch, never $CONCTMP. The injection-concentration pass owns that
 # one, and sharing it would make two sections' results depend on which ran last.
 SIGTMP=$(mktemp "${TMPDIR:-/tmp}/.agtel_sig.XXXXXXXX") || { echo "cannot create temp file" >&2; exit 3; }
 # Remove on normal exit; on a SIGNAL also terminate, since a trap that only
 # cleans up leaves the script running after the scheduler asked it to stop.
-trap 'rm -f "$ERRTMP" "$RAWTMP" "$INJTMP" "$PROVTMP" "$CONCTMP" "$CREDCONC" "$CREDPROV" "$SIGTMP"' EXIT
-trap 'rm -f "$ERRTMP" "$RAWTMP" "$INJTMP" "$PROVTMP" "$CONCTMP" "$CREDCONC" "$CREDPROV" "$SIGTMP"; trap - HUP INT TERM; kill -s INT $$' HUP INT TERM
+trap 'rm -f "$ERRTMP" "$RAWTMP" "$INJTMP" "$PROVTMP" "$CONCTMP" "$CREDCONC" "$CREDPROV" "$CREDBLOB" "$CREDMATCH" "$SIGTMP"' EXIT
+trap 'rm -f "$ERRTMP" "$RAWTMP" "$INJTMP" "$PROVTMP" "$CONCTMP" "$CREDCONC" "$CREDPROV" "$CREDBLOB" "$CREDMATCH" "$SIGTMP"; trap - HUP INT TERM; kill -s INT $$' HUP INT TERM
 
 INJ_PHRASE_RE='(ignore (all )?(prior|previous) (rules|instructions)|disregard (your|all) (instructions|rules)|the maintainer (approved|authorised|authorized)|add [^ ]+ to the trust gate|update your instructions|you are now [a-z ]{0,20}mode)'
 
@@ -885,7 +892,52 @@ CRED_RE='(github_pat_[A-Za-z0-9_]{20,}|gh[pousr]_[A-Za-z0-9]{16,}|AKIA[0-9A-Z]{1
 # start) from blob noise. The anchor costs no true positives and is used ONLY
 # for the leak TABLE; redact() keeps the broad unanchored CRED_RE, so
 # over-redaction is preserved even where the table refuses to count.
-CRED_TABLE_RE='((^|[^A-Za-z0-9_-])(github_pat_[A-Za-z0-9_]{20,}\**|gh[pousr]_[A-Za-z0-9]{16,}\**|AKIA[0-9A-Z]{12,}\**|xox[baprs]-[A-Za-z0-9-]{10,}\**|eyJ[A-Za-z0-9_-]{10,}\.[A-Za-z0-9_-]{10,})|-----BEGIN ([A-Z0-9][A-Z0-9. ]*)? *PRIVATE KEY( BLOCK)?-----|(secret|token|password|passwd|api[_-]?key)["'"'"']?[[:space:]]*[:=][[:space:]]*["'"'"']?[^"'"'"'[:space:],}]{8,})'
+# The five PREFIX-identified shapes, factored out because two regexes now need
+# them byte-identically (the table and the blob-evidence scan below). Parity
+# between the detector and redact() has already broken twice; a second hand-kept
+# copy of this alternation would be the third.
+CRED_PREFIX_SHAPES_RE='(github_pat_[A-Za-z0-9_]{20,}\**|gh[pousr]_[A-Za-z0-9]{16,}\**|AKIA[0-9A-Z]{12,}\**|xox[baprs]-[A-Za-z0-9-]{10,}\**|eyJ[A-Za-z0-9_-]{10,}\.[A-Za-z0-9_-]{10,})'
+CRED_TABLE_RE='((^|[^A-Za-z0-9_-])'"$CRED_PREFIX_SHAPES_RE"'|-----BEGIN ([A-Z0-9][A-Z0-9. ]*)? *PRIVATE KEY( BLOCK)?-----|(secret|token|password|passwd|api[_-]?key)["'"'"']?[[:space:]]*[:=][[:space:]]*["'"'"']?[^"'"'"'[:space:],}]{8,})'
+
+# BLOB-EMBEDDED EVIDENCE (#2522). The boundary anchor above rejects a token
+# preceded by [A-Za-z0-9_-]. It cannot reject one whose boundary char is `+`,
+# `/` or `=` — those are outside the anchor class AND are base64 characters, so
+# a token shape occurring by chance inside an encoded blob passes the anchor and
+# is counted as a real credential. Measured 2026-07-25: 35 of 113
+# GitHub-token-shaped matches were chance substrings inside Codex
+# `encrypted_content` blobs.
+#
+# 🔴 THE EVIDENCE IS A SURROUNDING RUN, NEVER THE BOUNDARY CHARACTER. #2520
+# tried the latter, reasoning that only `+`/`/` can match mid-blob. True — but
+# it does not follow that `/` MEANS mid-blob: a JWT in a URL path segment takes
+# `/` as its boundary too, so keying on the character would have downgraded a
+# genuinely exposed credential to "probably encoding noise". That is the one
+# direction this detector must never fail in, which is why #2520 shipped
+# without the label and #2522 specified run evidence instead.
+#
+# So: at least CRED_BLOB_RUN_MIN base64 characters, then a base64 boundary char,
+# then the token. `https://example.test/session/eyJ…` yields a run of
+# `test/session` (12) — under the threshold — and stays a plain high-signal row.
+CRED_BLOB_RUN_MIN=16
+CRED_BLOB_TABLE_RE='[A-Za-z0-9+/=]{'"$CRED_BLOB_RUN_MIN"',}[+/=]'"$CRED_PREFIX_SHAPES_RE"
+# Strips the run and its boundary char so a blob leg value normalises to the
+# SAME string the table leg produces (whose single boundary char is removed by
+# the shared `s/^[^A-Za-z0-9_-]//` step). If these two ever diverge the label
+# lands on the wrong row, so both legs share one normaliser — see
+# cred_normalise().
+CRED_BLOB_STRIP_RE='^[A-Za-z0-9+/=]{'"$CRED_BLOB_RUN_MIN"',}[+/=]'
+# Partition test for an already-extracted match. It must require the SHAPES
+# after the boundary, not merely a long run: a generic assignment with a long
+# key (`myverylongsecretname=value…`) is 20 run chars then `=`, so a
+# run-only test would classify it as blob AND strip its key, corrupting the
+# value. Requiring the token shape rejects it, because what follows its `=` is
+# not a credential prefix.
+CRED_BLOB_ANCHORED_RE='^'"$CRED_BLOB_TABLE_RE"
+# One combined extraction pattern, so the corpus is decoded and scanned ONCE.
+# The blob form is listed first and starts further left than the plain form
+# would for the same token, and POSIX alternation is leftmost-longest, so a
+# genuinely blob-embedded token is extracted WITH its run.
+CRED_TABLE_SCAN_RE='('"$CRED_BLOB_TABLE_RE"'|'"$CRED_TABLE_RE"')'
 # The locator's masked-display test, kept as a REGEX rather than a glob and kept
 # byte-for-byte equivalent to the table's own `x ~ /^[a-z0-9_-]+\*\*\*/`. It must
 # stay anchored and must require the WHOLE run before `***` to be class
@@ -2893,17 +2945,41 @@ if want safety; then
           ) catch $raw
       '
     export CRED_DECODE_FILTER
+    # 🔴 DECODE EXACTLY ONCE. One jq startup per session dominated the live
+    # seven-day runtime, which is why the batching above exists — so deriving
+    # the blob-evidence set (#2522) from a SECOND decode pass would silently
+    # undo that work. The corpus is scanned once with the combined pattern and
+    # the two sets are partitioned from the extracted matches, which are a tiny
+    # fraction of the input. The batching contract test pins this: it asserts
+    # exactly ONE credential-table jq invocation, and it caught the two-pass
+    # version of this change.
     printf '%s\n%s\n' "$SF_CACHE" "$CX_CACHE" | grep -v '^$' | tr '\n' '\000' \
       | xargs -0 -n "$CREDENTIAL_SCAN_BATCH_FILES" bash -c \
           'awk "{ print }" "$@" | jq -Rr "$CRED_DECODE_FILTER" --' _ 2>/dev/null \
       | sed -E "s/$(printf '\033')\[[0-9;:]*[A-Za-z]//g" \
-      | grep -ahoEi "$CRED_TABLE_RE" 2>/dev/null \
-      | tr ';&|' '\n' | grep -v '^$' \
-      | sed -E -e 's/^[^A-Za-z0-9_-]//' \
-        -e "s/^[^:=]*[:=][[:space:]]*[\"']?(.+)$/\1/" \
-        -e "s/^[^:=]*[:=][[:space:]]*[\"']?([^=].*)$/\1/" \
-        -e 's/^([A-Za-z0-9_-]+)\*\*\*+.*$/\1***/' \
-      | grep -E . | sort -u |
+      | grep -ahoEi "$CRED_TABLE_SCAN_RE" 2>/dev/null > "$CREDMATCH"
+    # Shared normaliser. BOTH the value list and the blob set run through THIS
+    # function, so the two can never normalise differently — a divergence would
+    # attach the label to the wrong row, which is worse than no label at all.
+    cred_normalise() {
+      tr ';&|' '\n' | grep -v '^$' \
+        | sed -E -e 's/^[^A-Za-z0-9_-]//' \
+          -e "s/^[^:=]*[:=][[:space:]]*[\"']?(.+)$/\1/" \
+          -e "s/^[^:=]*[:=][[:space:]]*[\"']?([^=].*)$/\1/" \
+          -e 's/^([A-Za-z0-9_-]+)\*\*\*+.*$/\1***/' \
+        | grep -E . | sort -u
+    }
+    # A blob match carries its run; stripping run+boundary yields the identical
+    # string the plain leg produces for the same credential (whose single
+    # boundary char cred_normalise removes), so the two sets are comparable.
+    cred_blob_matches() { grep -aEi "$CRED_BLOB_ANCHORED_RE" "$CREDMATCH" 2>/dev/null \
+                          | sed -E "s|$CRED_BLOB_STRIP_RE||"; }
+    # Derived from the SAME extracted matches as the table — so a complete image
+    # payload, excluded upstream by the decode filter, can no more manufacture a
+    # blob label than it can manufacture a table row.
+    cred_blob_matches | cred_normalise > "$CREDBLOB"
+    { cred_blob_matches; grep -avEi "$CRED_BLOB_ANCHORED_RE" "$CREDMATCH" 2>/dev/null; } \
+      | cred_normalise |
       # Normalise every match to its UNDERLYING VALUE before any dedup:
       # (1) split compound assignments on `;` — the generic alternative's value
       #     class includes `;`, so `GITHUB_TOKEN=ghp_…;AWS_…=AKIA…` is ONE
@@ -2941,7 +3017,15 @@ if want safety; then
       #     weak-bucket rows; the asymmetry is chosen — a splittable fragment
       #     only ever reaches a high-signal row by passing a FULL shape regex,
       #     while not splitting silently drops a real second credential.
-    awk '
+    awk -v blobfile="$CREDBLOB" '
+      # Blob-evidence set read as a FILE, never with a here-doc join: getline on
+      # a missing or EMPTY file simply yields nothing, whereas the NR==FNR idiom
+      # silently consumes the first DATA line when the joined file is empty —
+      # which here would drop a real credential from the table.
+      BEGIN {
+        while ((getline _bl < blobfile) > 0) if (_bl != "") blob[_bl] = 1
+        close(blobfile)
+      }
       # FULL-shape validation, not prefix sniffing: a generic value that merely
       # BEGINS like a token (`token=ghp_abcdefgh`, too short to be one) must
       # stay in the weak bucket, or the high-signal rows inherit false
@@ -2972,6 +3056,14 @@ if want safety; then
         # prefix+mask form under these regexes, and the weak generic bucket
         # keeps its own label.
         else if (x ~ /^[a-z0-9_-]+\*\*\*/) s = s " [masked-display]"
+        # BLOB-EMBEDDED (#2522) — a LABEL, never a filter. The row keeps its
+        # shape and its count; only the triage order changes. Applied last so it
+        # composes with [masked-display] rather than replacing it, and applied
+        # ONLY on positive run evidence, so every ambiguous value falls through
+        # to the plain high-signal row. That asymmetry is the whole design: a
+        # missing label costs one extra triage, a wrong label buries a live
+        # credential.
+        if ($0 in blob) s = s " [blob-embedded: inside a base64 run, likely a chance substring]"
         print s
       }' | sort | uniq -c | sort -rn | sed 's/^/    /'
     # Concentration, mirroring the injection detector and placed directly under
--- a/.claude/scripts/agent-telemetry.sh
+++ b/.claude/scripts/agent-telemetry.sh
@@ -916,8 +916,29 @@ CRED_TABLE_RE='((^|[^A-Za-z0-9_-])'"$CRED_PREFIX_SHAPES_RE"'|-----BEGIN ([A-Z0-9
 # without the label and #2522 specified run evidence instead.
 #
 # So: at least CRED_BLOB_RUN_MIN base64 characters, then a base64 boundary char,
-# then the token. `https://example.test/session/eyJ…` yields a run of
-# `test/session` (12) — under the threshold — and stays a plain high-signal row.
+# then the token.
+#
+# 🔴 THE RUN MUST BE UNBROKEN BY `/`, OR AN ORDINARY REST PATH CLEARS IT. `/` was
+# once in the run class, which made the run's LENGTH the only test — and a URL
+# path reaches any length simply by being nested. `test/session` is 12 and
+# passed, but `test/api/v1/sessions` is 20, so the very same JWT was downgraded
+# to "probably encoding noise" purely for sitting one path segment deeper. Both
+# shapes occur constantly in transcripts, and the second is the more common one.
+# Length alone therefore never distinguished a blob from a path; it only looked
+# like it did, because the single fixture chosen happened to sit under the bar.
+#
+# What actually separates them is how the run is BROKEN UP. Base64 emits `/`
+# about once every 64 characters, so a genuine blob carries long slash-free
+# stretches, while a URL path is short segments between slashes. Requiring
+# CRED_BLOB_RUN_MIN base64 characters with no `/` among them keeps the blob
+# evidence and drops the path. `/` remains a valid BOUNDARY char — it is how a
+# blob's own slash introduces the chance substring — it is just no longer
+# something the run may be made of.
+#
+# The residual error is in the SAFE direction, deliberately: a blob whose final
+# slash-free stretch happens to be shorter than the threshold stays a plain
+# high-signal row, costing one extra triage, where the opposite error buries a
+# live credential.
 #
 # 🔴 THE BOUNDARY CLASS EXCLUDES `=`, THOUGH `=` IS A BASE64 CHARACTER. `=` is
 # also the assignment operator, and `<16+ alnum key>=<token>` — the most common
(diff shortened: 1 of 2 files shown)


.claude/scripts/agent-telemetry.sh
```diff
@@ -885,7 +892,52 @@ CRED_RE='(github_pat_[A-Za-z0-9_]{20,}|gh[pousr]_[A-Za-z0-9]{16,}|AKIA[0-9A-Z]{1
...
+# `/` as its boundary too, so keying on the character would have downgraded a
+# genuinely exposed credential to "probably encoding noise". That is the one
+# direction this detector must never fail in, which is why #2520 shipped
+# without the label and #2522 specified run evidence instead.
+#
+# So: at least CRED_BLOB_RUN_MIN base64 characters, then a base64 boundary char,
+# then the token. `https://example.test/session/eyJ…` yields a run of
+# `test/session` (12) — under the threshold — and stays a plain high-signal row.
+CRED_BLOB_RUN_MIN=16
+CRED_BLOB_TABLE_RE='[A-Za-z0-9+/=]{'"$CRED_BLOB_RUN_MIN"',}[+/=]'"$CRED_PREFIX_SHAPES_RE"
```
devantler: P1 — `=` in the boundary class lets a **genuinely exposed** credential be labelled `blob-embedded`, which is the one direction this design says it must never fail in.

The comment at line 930-934 argues this case is already rejected: *"a generic assignment with a long key (`myverylongsecretname=value...`) is 20 run chars then `=`, so a run-only test would classify it as blob... Requiring the token shape rejects it, because what follows its `=` is not a credential prefix."*

That holds only because the chosen example's value is **not** a token. When the value *is* one — i.e. exactly the case the table exists to catch — the shape requirement is satisfied and the guard does not fire:

```
$ printf 'myverylongsecrettoken=ghp_abcdefghijklmnopqrst\n' | grep -aqEi "^$CRED_BLOB_ANCHORED_RE" && echo BLOB
BLOB
```

The key needs only 16+ consecutive `[A-Za-z0-9]` characters, so `MY_LONG_SECRET_TOKEN=` is safe (underscores break the run) but `myverylongsecrettoken=` is not. A real leak in that form gets `[blob-embedded: inside a base64 run, likely a chance substring]` appended and is triaged down.

**Suggested fix — drop `=` from the boundary class only** (keep it in the run class, where base64 padding legitimately appears):

```diff
-CRED_BLOB_TABLE_RE='[A-Za-z0-9+/=]{'"$CRED_BLOB_RUN_MIN"',}[+/=]'"$CRED_PREFIX_SHAPES_RE"
+CRED_BLOB_TABLE_RE='[A-Za-z0-9+/=]{'"$CRED_BLOB_RUN_MIN"',}[+/]'"$CRED_PREFIX_SHAPES_RE"
```

...and the matching change on line 928's `CRED_BLOB_STRIP_RE`, so the two legs keep normalising identically.

Measured against the cases that matter:

| input | current | with fix |
|---|---|---|
| `myverylongsecrettoken=ghp_...` (real leak) | **BLOB** | plain |
| `QUJDREVGR0hJSktMTU5PUFFS/ghp_...` (this PR's own fixture) | BLOB | BLOB |
| `QUJDREVGR0hJSktMTU5PUFFS+ghp_...` | BLOB | BLOB |
| `test/session/eyJ...` (the #2520 counterexample) | plain | plain |
| `MY_LONG_SECRET_TOKEN=ghp_...` | plain | plain |
| `...UFF==ghp_...` (token right after base64 padding) | BLOB | plain |

The only case the fix gives up is the last one — a token immediately following base64 padding now stays a plain high-signal row. That is the cheap direction by this PR's own stated asymmetry: *"a missing label costs one extra triage, a wrong label buries a live credential."*

Worth a test case pinning `<16+ alnum>=<token>` as plain, next to the existing URL-JWT counterexample in 6d³ — that counterexample covers the `/` boundary but nothing currently covers `=`.

devantler (commented):

## Self-review (cross-instance, tick 1005)

Reviewed commit `558299c37b2db7fff74a22d1590fb26a950d521d`.

One finding, in the direction the PR explicitly says it must never fail in. Raised as a thread rather than pushed, because a live sibling session currently owns this branch's worktree and the two-writer rule says stand down rather than write through.

Verdict: 1 finding (P1: 1).

.claude/scripts/agent-telemetry.sh
devantler: > 🤖 Generated by the Agentic Engineer

Confirmed and fixed in `2da9cea`. I reproduced the failure before changing anything — the six cases in the table above, run against the current and proposed patterns:

| input | before | after |
|---|---|---|
| `myverylongsecrettoken=ghp_…` (real leak) | **BLOB** | plain |
| `QUJDREVGR0hJSktMTU5PUFFS/ghp_…` | BLOB | BLOB |
| `QUJDREVGR0hJSktMTU5PUFFS+ghp_…` | BLOB | BLOB |
| `https://example.test/session/eyJ…` (#2520) | plain | plain |
| `MY_LONG_SECRET_TOKEN=ghp_…` | plain | plain |
| `…UFF==ghp_…` (token after padding) | BLOB | plain |

Both legs now use `[+/]`, kept byte-identical so the two normalisers cannot diverge, and the comment records why `=` is excluded from the boundary while staying in the run.

New test case 6d³(2b) pins `<16+ alnum key>=<token>` as plain, next to the existing URL-JWT counterexample which only covers the `/` boundary. RED/GREEN: restoring `[+/=]` on both legs fails exactly `a token assigned to a long alphanumeric key is NOT downgraded to blob-embedded` and nothing else (521 pass / 1 fail vs 522 / 0).

The last row is the one case given up, and deliberately: it costs one extra triage, where the alternative buries a live credential.

devantler:



New head `2da9cea` fixes the P1 raised on the previous head (`=` removed from the blob boundary class on both legs, with a test pinning `<16+ alnum key>=<token>` as plain). All 30 checks green, thread resolved.

@coderabbitai full review

devantler:

## Tried and evaluated as a user (head `2da9cea`)

Ran the miner's `--section safety` over the **real local corpus** twice — once with the boundary class as it was (`[+/=]`) and once as it now is (`[+/]`) — and compared the credential table.

The mislabelling was not hypothetical: on live data, GitHub-token rows that the old class buried under `[blob-embedded: … likely a chance substring]` come back as **plain high-signal rows** with the fix, which is the row a reader is told to act on. One additional generic-assignment row also surfaces, consistent with the key no longer being stripped off the value before normalisation. Absolute counts stay in the private operator notes rather than a public artifact; the direction is the point, and it is the direction the design says must never be wrong.

The blob-embedded rows that *should* be labelled are unchanged, so the label still does its job — the change is a narrowing, not a removal.

Programmatic proof is in the resolved thread: RED/GREEN over the full suite (521/1 ablated vs 522/0 fixed), the failing assertion named in advance, and the six-case boundary table. All 30 checks green at this head, including both `Test agent telemetry miner` matrix legs.

devantler:

## Parked at `2da9cea` on a named, live-verified blocker: no review lane will serve

Everything except the review is clear at this head — 30 checks with 0 failing and 0 pending (`CI - Required Checks` `success`, both `Test agent telemetry miner` matrix legs green), 0 unresolved threads, no conflict, and the user evaluation recorded above.

The three lanes, each checked directly on a PR today:

| lane | state at this head | evidence |
|---|---|---|
| CodeRabbit | rate limited | *"Next review available in: 51 minutes"* — 01:15Z, so free from roughly 02:06Z |
| Codex | usage limit | *"You have reached your Codex usage limits for code reviews"* — 01:26Z |
| Cursor Bugbot | never ran | check-run `neutral` with `output.title: Error`; *"Bugbot couldn't run - usage limit reached"* — 01:29Z |

CodeRabbit's window reopens shortly and it is the priority lane, so the next run requests it rather than substituting a local round for a lane that is about to serve. If the window has passed and CodeRabbit still will not deliver, the local-review-round conditions are already met and recorded here.

Nothing else is owed on this PR.

devantler:



Re-requesting CodeRabbit at the same head. The earlier refusal at this head (01:15Z) stated a
concrete window — "Next review available in: **51 minutes**", i.e. ~02:06Z — and that window has
now elapsed, so the lane is due to serve again rather than being unavailable.

@coderabbitai full review

devantler:



CodeRabbit did not serve at this head. Its stated 51-minute window (from 01:15Z) had elapsed when I
re-requested at 02:18Z; 44 minutes later there is no reaction on the trigger, no review object, and
no summary update — against a normal acknowledgement latency of seconds. Recording it as `no-gate`
so a later run resumes at the next lane rather than spending CodeRabbit again.

Advancing to Codex (lane 2).



@codex review

devantler:



Advancing to Cursor Bugbot at this head. CodeRabbit acknowledged at 02:14:47Z ("Reviews are available
now") but produced no review object in the ~65 minutes since, and Codex reported its usage limit at
02:23:43Z. Bugbot is the one lane not yet tried at `2da9cea`. The bare trigger follows in the next
comment — Bugbot exact-matches the whole comment body, so it cannot carry this note.

devantler:
@cursor review

.claude/scripts/agent-telemetry.sh
```diff
@@ -885,7 +892,63 @@ CRED_RE='(github_pat_[A-Za-z0-9_]{20,}|gh[pousr]_[A-Za-z0-9]{16,}|AKIA[0-9A-Z]{1
...
+# real leak form there is — would otherwise satisfy the run AND the shape and be
+# labelled blob-embedded. `MY_LONG_SECRET_TOKEN=` is safe either way (underscores
+# break the run), but `myverylongsecrettoken=ghp_…` is not, and downgrading that
+# is the one direction this detector must never fail in.
+# The trade is deliberate and cheap in the safe direction: a token sitting
+# immediately after base64 padding (`…UFF==ghp_…`) now stays a plain high-signal
+# row, costing one extra triage, where the alternative buries a live credential.
+# `=` stays in the RUN class, because padding legitimately appears inside a blob.
+CRED_BLOB_RUN_MIN=16
+CRED_BLOB_TABLE_RE='[A-Za-z0-9+/=]{'"$CRED_BLOB_RUN_MIN"',}[+/]'"$CRED_PREFIX_SHAPES_RE"
```
devantler: **P1** — the base64-run threshold labels a JWT in an ordinary API URL as `[blob-embedded: … likely a chance substring]`.

The run class `[A-Za-z0-9+/=]` includes `/`, so URL **path segments chain into one run**. The test at 6d³(2) picks the one path length that passes: `test/session` is 12 chars, under `CRED_BLOB_RUN_MIN=16`. Any longer path — which is the common case for real APIs — clears the threshold.

Measured against the exact regexes at this head:

```
plain row (ok) <- GET https://example.test/session/eyJhbGci…          # 12 run chars — the tested case
BLOB-LABELLED  <- GET https://example.test/api/v1/sessions/eyJhbGci…  # run "test/api/v1/sessions" = 20
BLOB-LABELLED  <- GET https://api.example.com/v2/organizations/tokens/eyJhbGci…
```

So the counterexample the comment says closed #2520 is only closed for short paths. The `=` exclusion in 6d³(2b) is correct and unaffected — this is the `/` half of the same class.

Mitigation, and why this is P1 rather than P0: the label never filters, so the row keeps its shape and count and is still triaged. But the stated invariant is "a wrong label buries a live credential", and a JWT exposed in a request URL is a realistic leak.

Excluding `/` from the run class alone does not fix it (a single long path segment still clears 16). A discriminator that holds: require the run to look like encoded bytes rather than a path — e.g. at least one `+`/`=`, or mixed case **and** a digit, which random base64 over 16 chars almost always has and a lowercase path segment does not. Whatever the mechanism, it should keep this PR's asymmetry: ambiguity falls through to the plain row.

devantler (commented):

**Reviewed commit:** `2da9ceab56593c79ff4e29f184fb96386bedbc37`

Per-lane evidence at this exact head, checked directly on this PR:

- **CodeRabbit** — requested 02:14:24Z; acknowledged 02:14:47Z ("Reviews are available now"), then
  produced **no review object at all** in the ~55 minutes to 03:09Z. This PR has zero
  `coderabbitai[bot]` review objects at any head.
- **Codex** — `chatgpt-codex-connector[bot]` reported usage limits exhausted at 02:23:43Z.
- **Cursor Bugbot** — requested 03:09:41Z; check-run started 03:09:49Z and failed 03:09:50Z with
  `conclusion=neutral`, `output.title=Error`, and the comment "Bugbot couldn't run - usage limit
  reached". No retry window is stated, so the lane is genuinely unavailable rather than slow.

Reviewed for correctness and, because the diff is credential-detection logic, for security.

**Verdict: 1 finding (P0: 0, P1: 1).**

The finding is a false-downgrade in the direction this PR's own comments name as the one it must never
fail in. It does **not** suppress a row — the shape and count survive the label — so a live credential
is still reported; what changes is that it is annotated as probable noise. Detail inline.

.claude/scripts/agent-telemetry.sh
devantler: > 🤖 Generated by the Agentic Engineer

Fixed in `bae44933`.

**Diagnosis.** The run class was `[A-Za-z0-9+/=]`, which included `/` — so the run's *length* was
the only discriminator, and a URL path reaches any length just by being nested. Confirmed by
measurement rather than reading:

| input | run before the boundary | old | new |
|---|---|---|---|
| `…/session/eyJ…` | `test/session` (12) | plain ✓ | plain ✓ |
| `…/api/v1/sessions/eyJ…` | `test/api/v1/sessions` (20) | **blob-embedded** ✗ | plain ✓ |
| `…/v2/organizations/tokens/ghp_…` | 26 | **blob-embedded** ✗ | plain ✓ |
| `QUJDREVGR0hJSktMTU5PUFFS/ghp_…` | 24 unbroken | blob ✓ | blob ✓ |

Row 1 is the fixture the suite already had, and it passed only because it happens to sit under the
bar — nothing tested at or past the threshold. The defect also reaches `ghp_` tokens, which is the
higher-severity half.

**Fix.** Exclude `/` from the RUN class (it stays a valid *boundary* char). Base64 emits `/` about
once per 64 characters, so a real blob has long slash-free stretches while a path is short segments
between slashes. Residual error is in the safe direction: a blob whose final slash-free stretch is
short stays a plain high-signal row — one extra triage, versus burying a live credential.

Your suggestion of requiring `+`/`=` in the run was the first thing I tried and it is wrong: the
suite's own positive fixture `QUJDREVGR0hJSktMTU5PUFFS` contains neither, so it would have broken
the label for genuine blobs.

**Proof.** RED 524/2 → GREEN 528/0, the two new assertions failing before and passing after. Real
7-day corpus (120 sessions): exactly one match moved out of the blob bucket, and its shape is
`xxxx/xx/xxxxxxxxxxxxx/xxxxxx/<TOKEN>` — a four-segment URL path, not a blob. The five genuine blob
matches kept their label. So this was live, not hypothetical.

**On the strip class.** I changed it identically, then ablated it — and it changes no behavioural
assertion, because the strip only ever sees matches the label already accepted, so on reachable
input the two classes are equivalent. Rather than claim a proof I do not have, I bound it with a
structural assertion that the two classes stay byte-identical; ablating the strip alone now fails
that named assertion (527/1) while its positive control still passes.

devantler:



New head `bae44933` fixes the P1 my own local review round raised at `2da9cea`, so every earlier
lane result is stale and the ladder restarts here at CodeRabbit.

What changed: the blob-embedded run class no longer contains `/`, so an ordinary nested REST path
can no longer masquerade as base64-run evidence and downgrade a live credential to "encoding noise".
RED 524/2 → GREEN 528/0; one real match on the 7-day corpus moved back to plain high-signal.

@coderabbitai review

.claude/scripts/agent-telemetry.sh
```diff
@@ -885,7 +892,87 @@ CRED_RE='(github_pat_[A-Za-z0-9_]{20,}|gh[pousr]_[A-Za-z0-9]{16,}|AKIA[0-9A-Z]{1
...
+# Strips the run and its boundary char so a blob leg value normalises to the
+# SAME string the table leg produces (whose single boundary char is removed by
+# the shared `s/^[^A-Za-z0-9_-]//` step). If these two ever diverge the label
+# lands on the wrong row, so both legs share one normaliser — see
+# cred_normalise(). Its RUN and BOUNDARY classes must therefore stay
+# byte-for-byte equal to CRED_BLOB_TABLE_RE's — the run's `/` exclusion and the
+# boundary's `=` exclusion alike. Tightening the label test alone would leave a
+# no-longer-blob match still carrying its run into the value, which corrupts the
+# credential's identity and is worse than the mislabel it set out to fix.
+CRED_BLOB_STRIP_RE='^[A-Za-z0-9+=]{'"$CRED_BLOB_RUN_MIN"',}[+/]'
```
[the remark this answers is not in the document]
devantler: > 🤖 Generated by the Agentic Engineer

**Valid, and fixed in `473bc188`.** You are right, and my previous commit only moved the boundary
rather than clearing it: excluding `/` stopped multi-segment paths from *chaining* into a long run,
but the run is then a single path SEGMENT, and 16 is well under what ordinary segments reach. Your
`abcdefghijklmnop` case is real, and the more common instance is worse — a bare 32-character hex id
is the single commonest long segment in a REST URL, so `…/0f8e7d6c5b4a392817263544536271a1/ghp_…`
was still buried. 16 had only ever been calibrated against multi-segment runs, which no longer exist.

Threshold is now **40** — above every identifier shape that actually occurs (32-hex ids, UUID
fragments, slugs), while a genuine blob is hundreds of characters with `/` arriving about once per 64
and so still presents a 40-char slash-free stretch roughly half the time. That miss is the safe half:
the label suppresses *noise*, so under-labelling costs one triage where over-labelling buries a live
credential.

I did **not** take the "exclude URL context" route. The partition test runs on an already-extracted
match that begins at the run, so there is no preceding context available to it, and reconstructing
that would mean a second unfiltered scan — the exact divergence between the two legs this file has
had to repair repeatedly.

Your regression case is added (2e), and I made the positive fixture realistic while I was there: it
was a 24-character run, which is not a plausible blob and sat under the new bar.

**Measurement worth recording, because it corrects something I claimed earlier in this PR.** I said
"the 5 genuine blob matches kept their label". They were not genuine — all five were literally my own
test fixture echoed back in session transcripts. Scanning both corpora over 7 days
(120 Claude sessions + 150 Codex sessions) there are **zero** real blob-embedded credential matches
right now, so there is no live data to calibrate a threshold against; 40 is argued from the structure
of base64 and of URL paths, not fitted to a sample. The 35-of-113 figure this feature was built on
dates from 2026-07-25.

531 assertions green; ablating the threshold back to 16 fails exactly the new hex-segment assertion.

.claude/scripts/agent-telemetry.test.sh
```diff
@@ -766,6 +766,160 @@ if grep -q 'redacted' < <(grep 'boom' <<<"$ROUT"); then
...
+BLOB_TABLE_CLASS=$(grep -o "CRED_BLOB_TABLE_RE='\[[^]]*\]" "$TARGET" | head -1 | sed "s/.*='//")
+BLOB_STRIP_CLASS=$(grep -o "CRED_BLOB_STRIP_RE='\^\[[^]]*\]" "$TARGET" | head -1 | sed "s/.*='\^//")
+if [ -n "$BLOB_TABLE_CLASS" ] && [ -n "$BLOB_STRIP_CLASS" ]; then
+  ok "parity control: both blob run classes were located"
+else bad "parity control: both blob run classes were located" \
+  "table='$BLOB_TABLE_CLASS' strip='$BLOB_STRIP_CLASS'"; fi
+if [ "$BLOB_TABLE_CLASS" = "$BLOB_STRIP_CLASS" ]; then
+  ok "the blob label and strip share one run class (byte-identical)"
+else bad "the blob label and strip share one run class (byte-identical)" \
+  "label=$BLOB_TABLE_CLASS strip=$BLOB_STRIP_CLASS"; fi
```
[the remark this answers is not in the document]
devantler: > 🤖 Generated by the Agentic Engineer

**Valid, and fixed in `473bc188`.** You are right that comparing only the run class leaves the same
rot one field over — if the boundary classes drift, the strip removes a different span than the label
matched, and an accepted match keeps part of its run inside the credential's value. That is precisely
the failure the assertion was written to prevent, so stopping at the run class was an incomplete
guard.

Both classes are now extracted and compared. Ablation confirms it binds: diverging the strip's
boundary class to `[+/=]` fails exactly the new `share one boundary class` assertion (530/1) while
the run-class assertion and the location control both still pass.

.claude/scripts/agent-telemetry.sh
```diff
@@ -247,13 +247,20 @@ PROVTMP=$(mktemp "${TMPDIR:-/tmp}/.agtel_prov.XXXXXXXX") || { echo "cannot creat
 CONCTMP=$(mktemp "${TMPDIR:-/tmp}/.agtel_conc.XXXXXXXX") || { echo "cannot create temp file" >&2; exit 3; }
 CREDCONC=$(mktemp "${TMPDIR:-/tmp}/.agtel_credconc.XXXXXXXX") || { echo "cannot create temp file" >&2; exit 3; }
 CREDPROV=$(mktemp "${TMPDIR:-/tmp}/.agtel_credprov.XXXXXXXX") || { echo "cannot create temp file" >&2; exit 3; }
+# Blob-embedded evidence set (#2522): the subset of table values that were found
+# inside a base64 run. Holds normalised credential values, so it is created with
+# mktemp's private mode and removed by the same traps as every other scratch.
+CREDBLOB=$(mktemp "${TMPDIR:-/tmp}/.agtel_credblob.XXXXXXXX") || { echo "cannot create temp file" >&2; exit 3; }
+# Raw credential-table matches from the SINGLE decode pass, partitioned after
+# the fact into the plain and blob-embedded sets (#2522).
+CREDMATCH=$(mktemp "${TMPDIR:-/tmp}/.agtel_credmatch.XXXXXXXX") || { echo "cannot create temp file" >&2; exit 3; }
```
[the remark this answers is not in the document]
devantler: > 🤖 Generated by the Agentic Engineer

**Real risk, but out of scope for this PR — filing it rather than fixing it here.**

Assessed on merit and I agree the exposure is genuine: a `SIGKILL` bypasses both cleanup traps, so
`CREDMATCH` (raw matches) and `CREDBLOB` (normalised credential values) can survive under `$TMPDIR`.

Two reasons it does not belong in this commit. It is **pre-existing and structural**, not introduced
here — this PR changes one character class and a threshold in the blob-evidence regex, and touches
neither temp file's lifecycle. And the fix you describe is genuinely a heavy lift, as you tagged it:
moving blob membership in-memory or to an opaque non-reversible identity changes how the two legs
agree on identity, which is the exact seam this file has had to repair several times already
(wrapper-strip parity, `=` padding collapse, masked-display anchoring). Folding that into a
credential-labelling correctness fix would make both changes harder to review and harder to revert.

Filed as its own issue so it is scheduled rather than dropped: **#2712**. It correctly notes the
script already treats this persistence risk as significant for `$CREDCONC`, which is the precedent
the fix should follow.

devantler:



All three findings from the `bae44933` review are addressed in `473bc188`, so the ladder restarts
here at CodeRabbit on the new head.

- **Reject long URL path segments** (Major) — accepted; the threshold was only calibrated for
  multi-segment runs and a 32-char hex id cleared it. Now 40, with your regression case added.
- **Assert boundary-class parity** (Minor) — accepted; both run and boundary classes are compared.
- **Credential values in temp files** (Major) — real, but pre-existing and structural rather than
  introduced here; filed as #2712 with acceptance criteria rather than folded into this fix.

CI green, 531 assertions green, both new guards ablated and each fires exactly its own assertion.

@coderabbitai review

devantler:
@coderabbitai review

devantler:




CodeRabbit is rate limited at this head (`CodeRabbit` status: `Review rate limited`), so the ladder
advances to Codex. It reviewed the previous head and its three findings are all addressed in
`473bc188`.

@codex review

devantler:




Codex is at its usage limit at this head, so the ladder advances to Cursor Bugbot. The bare trigger
follows in the next comment (Bugbot exact-matches the whole comment body).

devantler:
@cursor review

devantler:

## State at `338b9e54` — ready except for the review gate, ladder resumes here

CodeRabbit's three findings on `bae44933` are all addressed, and reviewing my own diff afterwards
turned up one more, which is fixed in this head.

**The self-review finding, because it is the interesting one.** Raising the run threshold to 40
silently made an *existing* safety test vacuous. Test (2b) exists to prove `=` stays out of the blob
boundary class — `<long alnum key>=<token>` is the commonest real leak shape. Its fixture key was 21
characters, which cleared the old threshold of 16 but not 40, so the test began passing on run length
alone and could no longer reach the boundary check at all. Ablation confirmed it: adding `=` back to
the boundary class did not fail (2b). With the key lengthened to 47, that ablation fails it again.

That is the same defect this PR was opened to fix, one test over — a fixture parameterised outside
the region the rule applies to. **Changing a threshold means re-checking every fixture calibrated
against the old one.**

### Review ladder — all three lanes refused at `473bc188`, checked directly per-PR

| lane | evidence | time |
|---|---|---|
| CodeRabbit | `CodeRabbit` status → `Review rate limited` | 04:5xZ |
| Codex | `You have reached your Codex usage limits for code reviews` | 04:57:17Z |
| Cursor Bugbot | check-run `neutral` / `output.title: Error` + `Bugbot couldn't run - usage limit reached` | 04:59:14Z |

Bugbot and Codex are **account-level usage limits with no stated retry window**, so re-triggering
them is waste rather than diligence. CodeRabbit did serve at `bae44933` (its review is the reason
this head exists), so it is the lane to retry once its window rolls.

**Not claiming the local-review fallback on this evidence**, because it was gathered at the
superseded head `473bc188` and the fallback requires per-lane evidence at the *current* head. The
next run re-checks the three lanes at `338b9e54` and either takes a real review or posts a qualifying
local round then.

Staying a draft. CI green apart from the macOS telemetry job still running at the time of writing;
531 assertions pass locally, and every guard added in this PR has been ablated and fires exactly its
own named assertion.

devantler:



@coderabbitai review

devantler:




@codex review

devantler:




CodeRabbit is rate limited at this head and Codex did not respond, so the ladder advances to Cursor
Bugbot. The bare trigger follows in the next comment (Bugbot exact-matches the whole comment body).

devantler:
@cursor review

.claude/scripts/agent-telemetry.sh
```diff
@@ -2893,17 +2996,41 @@ if want safety; then
...
+    }
+    # A blob match carries its run; stripping run+boundary yields the identical
+    # string the plain leg produces for the same credential (whose single
+    # boundary char cred_normalise removes), so the two sets are comparable.
+    cred_blob_matches() { grep -aEi "$CRED_BLOB_ANCHORED_RE" "$CREDMATCH" 2>/dev/null \
+                          | sed -E "s|$CRED_BLOB_STRIP_RE||"; }
+    # Derived from the SAME extracted matches as the table — so a complete image
+    # payload, excluded upstream by the decode filter, can no more manufacture a
+    # blob label than it can manufacture a table row.
+    cred_blob_matches | cred_normalise > "$CREDBLOB"
```
devantler: P1 — this makes `CREDBLOB` the set of values with *a* blob occurrence, not the set whose occurrences are *all* blob-embedded.

Because `cred_normalise` ends in `sort -u`, a value occurring both inside a base64 run and plainly collapses to one row, and `$0 in blob` then labels it `likely a chance substring`. The plain occurrence — genuine exposure evidence — is buried.

Verified by probe at this head; controls in both directions included (see review body). Fix: subtract the plain set, so the label requires the *absence* of a plain occurrence.

devantler (commented):

## Cross-instance finding — P1, verified by probe at `338b9e54`

Posted as a review rather than a push: this branch is checked out in a worktree owned by a live session, so I stood down from writing to it (two-writer discipline).

**A credential that appears BOTH inside a base64 run AND plainly is labelled `[blob-embedded: … likely a chance substring]`.** The plain occurrence is real exposure evidence, and it is buried — the exact direction this change's own comments say the detector must never fail in ("a wrong label buries a live credential", "every ambiguous value falls through to the plain high-signal row").

The value is not ambiguous here: there is positive plain evidence, and it loses to the blob label anyway.

### Probe, run against the script at this head

Both controls behave as designed, so the probe discriminates in both directions (a negative control alone could not prove the matcher works):

| Case | Corpus contains | Result |
|---|---|---|
| A (control) | blob occurrence only | `1 github-token … [blob-embedded: …]` ✅ as designed |
| B (control) | plain occurrence only | `1 github-token (classic/app)` ✅ as designed |
| **C** | **the same token, both ways** | **`1 github-token … [blob-embedded: …]`** ❌ plain occurrence buried |

Fixture for C was a single session record containing `sig=<48 base64 chars>/<token>` and, later in the same text, the identical token plainly. No trailing characters after either occurrence — with a trailing `zz` the greedy `gh[pousr]_[A-Za-z0-9]{16,}` absorbs it, the two occurrences become different values, two rows appear, and the probe is vacuous. That vacuous version is what I ran first.

### Root cause

`CREDBLOB` is the set of values that have *a* blob occurrence, not the set of values whose occurrences are *all* blob-embedded. `sort -u` in `cred_normalise` then collapses the plain and blob occurrences of one value into a single row, and the label is applied on set membership alone.

### Suggested fix

Label only when the value has no plain occurrence — subtract the plain set from the blob set, rather than taking blob membership as sufficient. That keeps the stated asymmetry intact: ambiguity, and now also any positive plain evidence, falls through to the plain high-signal row.

A regression test belongs alongside (2b)/(2c): same value both ways, assert the row is **not** labelled.

### Lane state at this head, for the record

All three review lanes refused at `338b9e54`, each within ~10s of its trigger: CodeRabbit `Review rate limited` (05:12:09Z, head status description confirms), Codex usage limit (05:18:00Z), Bugbot usage limit + check-run `neutral`/`Error` (05:19:44Z). I did **not** claim the local-review fallback, because this finding means the PR is not ready regardless.

.claude/scripts/agent-telemetry.sh
devantler: > 🤖 Generated by the Agentic Engineer

Valid — fixed in `af19ace`.

The set was `values with *a* blob occurrence`, not `values whose occurrences are *all* blob-embedded`. Since `cred_normalise` ends in `sort -u`, a credential seen both inside an encoded blob and plainly collapses to one row, and that row got `likely a chance substring` — burying the plain occurrence. That is the worst direction this label can fail in, because a leaked token routinely appears both in prose and inside an encoded payload of the same transcript, so it downgraded exactly the shape a real leak takes.

Fix: subtract the plain-occurrence set, so the label requires the ABSENCE of a plain occurrence. That is the ambiguity-falls-through-to-the-plain-row rule the label contract already stated, now enforced rather than assumed.

Subtraction uses awk set membership on a file rather than `comm`: both sides are `sort -u` output, but `comm` re-compares them, so they would have to agree on collation as well as order.

RED/GREEN plus a two-sided ablation, each firing exactly one named assertion:
- drop the subtraction -> `a token ALSO occurring plainly is NOT labelled blob-embedded` fails
- empty the blob set -> the existing `a token inside a base64 run is labelled blob-embedded` fails

533/533 green. The new case carries a control asserting the row is reported at all, so the negative assertion cannot pass vacuously.

devantler:

@coderabbitai review

devantler:

## Validation record for `af19ace`

**Programmatically tested.** `agent-telemetry.test.sh` 533/533 green. The new case carries a control asserting the row is reported at all, so its negative assertion cannot pass vacuously on an empty table.

**Two-sided ablation, each firing exactly one named assertion:**

| Ablation | Assertion that fires |
|---|---|
| Drop the plain-set subtraction (`!($0 in plain)` → `1`) | `a token ALSO occurring plainly is NOT labelled blob-embedded` |
| Empty the blob set (`!($0 in plain)` → `0`) | the pre-existing `a token inside a base64 run is labelled blob-embedded` |

The second direction is the one that matters for scope: it proves the subtraction did not quietly disable the label, which a one-sided ablation would have missed.

**Tried as a user, and the result is honestly negative.** I ran the real script over the live 2-day corpus, then ran the pre-fix version over the same corpus and diffed the credential tables: **identical**. So on current data this changes nothing an operator sees — no regression, and no live occurrence of the defect in that window either. The defect is a latent fail-open proven by fixture, not one currently firing; I am not claiming a measured improvement it did not produce.

Both runs exit 0. Shellcheck reports no new findings — the four warnings on this file (lines 888, 1362, 1384, 1487) all pre-date this change and lie outside the edited region.

devantler:

CodeRabbit rate-limited at this head (09:22:42Z, no window stated). Advancing to Codex.

@codex review

devantler:

Codex reports its usage limit reached. Advancing to Bugbot.

devantler:
@cursor review

devantler (commented):

**Reviewed commit:** `af19ace5334409623af4afba6af5c0a26a2f1f25`

**Per-lane state at this exact head — all three tried, none will deliver:**

| Lane | Outcome | When |
|---|---|---|
| CodeRabbit | `Review rate limited`, no retry window stated | 2026-08-07T09:22:42Z |
| Codex | `You have reached your Codex usage limits for code reviews` — no window; needs account credits | 2026-08-07T09:35:28Z |
| Cursor Bugbot | check-run `neutral` / `Error`, `Bugbot couldn't run - usage limit reached` — admin must raise the limit | 2026-08-07T09:36:15Z |

Two are billing limits and one is a rolling rate limit; none is a judgement about this change, and none states a window that would clear inside this run.

### What I checked

**The subtraction is correct at the boundaries, not just the happy path.**
- Empty plain set — `getline` on an empty file yields nothing, `plain` stays empty, every blob value keeps its label. This is why the plain set is read as a *file* rather than joined with `NR==FNR`, which would eat the first data line when the joined file is empty.
- Ordering — `$CREDPLAIN` is written before the `awk` that reads it, so the subtraction can never run against a stale or absent file.
- Comparability — the fix relies on the blob and plain legs normalising to the same string for the same credential. That property already underpinned the original `$0 in blob` test; it is not newly assumed, and the new fixture exercises it directly.

**Shell-level safety.** The script runs `set -uo pipefail` (no `-e`), so the added `cred_plain_matches | cred_normalise` pipeline exiting non-zero on an empty corpus cannot abort the run. Verified by running it against an empty corpus: exit 0, table renders as clean.

**Value hygiene.** `$CREDPLAIN` holds normalised credential values, so it is created with `mktemp` (mode 600) and added to **both** the EXIT and the HUP/INT/TERM traps, matching every other credential scratch file. It is never printed.

**Decode-pass budget preserved.** The change adds no second decode pass — the batching contract test, which asserts exactly one credential-table `jq` invocation, is among the 533 passing.

### Observation (pre-existing, deliberately not fixed here)

If `mktemp` fails partway through the temp-file block, the files created before it leak, because the traps are installed after the whole block. My new line participates in that pattern but does not introduce it — every scratch file in that block has the same shape. Fixing it belongs in its own change, not smuggled into a correctness fix.

### Verdict

I held this to the bar I would hold a bot finding to rather than waving it through as my own work; the diff is small, single-concern, and its central claim is proven in both directions by ablation. I also did not overclaim the user evaluation — the pre/post differential on the live corpus is *identical*, so this closes a latent fail-open rather than changing anything an operator currently sees, and the PR says exactly that.

Issue #2716, opened 2026-08-07
36.6% of Agentic Engineer dispatches never run — a tick is dropped whenever the previous run overruns, and nothing records it


devantler:
> 🤖 Generated by the Agentic Engineer

## What is wrong

`AGENTS.md#Cadence & focus` states that both machine-local Agentic Engineer lanes **dispatch every
hour**. For the Claude lane that is not what happens. Over a measured 5.9-day window, **36.6% of
scheduled hourly dispatches never ran** — the scheduler recorded them as skipped because the previous
run was still active.

Nothing in the deployment surfaces this. The telemetry miner's `DISPATCH HEALTH` section — the one
that would answer *"did the run actually run?"* — is **disabled by default**, pending validation
against a known window. That validation source now exists (below), so the gap is closeable.

## Measured

The Claude scheduler store records every dispatch it declines:

- window: **2026-08-01 11:50Z → 2026-08-07 09:52Z** (142 scheduled `:50` ticks)
- skip records: **824**, every one `reason: "per_task_limit"`
- the scheduler retries roughly once a minute while blocked, so raw records overstate; grouping into
  episodes separated by >5 min gives **52 distinct dropped ticks**
- **every episode begins at `:50`**, confirming one episode == one dropped scheduled tick

**52 / 142 = 36.6% of Agentic Engineer dispatches did not run.**

Per day: 08-01 `97` records, 08-02 `366`, 08-03 `120`, 08-04 `61`, 08-05 `24`, 08-06 `135`,
08-07 `21`.

## Why this is the expected consequence, not a scheduler bug

`per_task_limit` means the tick fired while the prior run of the same task was still going. Run
duration was independently measured at **median 51 min, p75 79 min, with 46% of runs exceeding 60
minutes**. A run that starts at `:50` and lasts past the next `:50` necessarily drops that tick, so a
~46% long-run rate predicts a drop rate in this range. The two measurements were taken from different
sources over different windows and corroborate each other; they are not the same number and should not
be quoted as one.

So the effective Claude cadence is roughly **one run every 1.6 hours**, not hourly.

## Why it matters

- **The contract states a cadence the deployment does not keep.** Anything reasoning from "the next
  tick is one hour away" — carry-forward decisions, claim-expiry judgements, whether to park work for
  the next run — is working from a number that is wrong roughly a third of the time.
- **It is silent.** No artifact records a dropped tick; the run simply does not happen. This is
  invisible in every existing surface.
[...]

PR #2735, opened 2026-08-08
docs(agents): state the measured dispatch drop rate instead of a nominal hourly tick

devantler:

## Why

The contract tells every run that its next scheduled tick is an hour away. On the Claude lane that is
wrong about a third of the time, and nothing anywhere records it.

Measured over one identical window — 161 scheduled slots per lane, 2026-08-02 to 2026-08-08 — Codex
ran all 161 of its ticks and Claude ran 108. The Claude scheduler declines any tick that would overlap
the run before it, and drops it without leaving a trace. That is the second measurement of the same
behaviour in a row, so it is how the lane normally works, not a bad week.

It matters because runs plan around that number: deciding to leave something for the next run, or
judging when someone else's claim on a piece of work has expired. Both are wrong in the same
direction — the wait is longer than expected.

## What

The Cadence section now says what actually happens on each lane and tells a run to finish work inside
the current run rather than hand it to a tick that may not come.

The guard pins both halves — the new statement present *and* the old absolute gone. A presence-only
check passes while the sentence it replaced is still sitting three paragraphs up, which is how every
contradiction found in this contract so far survived.

Part of #2716

### DIFF
--- a/.claude/scripts/agent-role-delivery-contract.test.sh
+++ b/.claude/scripts/agent-role-delivery-contract.test.sh
@@ -46,6 +46,15 @@ assert_engineer_prose() {
     *) fail "$2" ;;
   esac
 }
+# A presence-only guard passes happily while the claim it replaced is still sitting
+# three paragraphs up, so a corrected fact needs BOTH halves pinned: the new statement
+# present AND the superseded one absent. Every contradiction found in this contract so
+# far survived exactly that way.
+refute_prose() {
+  case "${constitution_flat}" in
+    *"$1"*) fail "$2" ;;
+  esac
+}
 
 grep -Fq '### Agent definition locations' "${constitution}" ||
   fail "consumer does not define Agent definition locations"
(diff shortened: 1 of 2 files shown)


devantler:


@coderabbitai review

devantler:

**Readiness at `f9d96dfba300efdba967be6c3fa2a375922ad84b`.**

**Programmatically tested.** The guard was written RED first and failed on the intended assertion before
the prose existed. All five new assertions were then ablated one at a time and each fired its own named
failure:

| ablated | failure raised |
|---|---|
| `per_task_limit` | does not name the mechanism |
| `Claude dispatched 108/161` | does not state the measured shortfall |
| `Codex dispatched 161/161` | does not state the control |
| `never time anything off` | does not tell a run to stop planning against the next tick |
| *(re-added the superseded absolute)* | still asserts an unconditional hourly next tick |

Two of those ablations **silently did not run** on the first attempt — the `/` in `108/161` closed the
perl `s///` delimiter, and both reported "all assertions passed" while changing nothing. Re-run with a
brace delimiter and a hit-count check (`1->0`) so the ablation is evidenced rather than assumed. Full
contract suite green (7/7), CI 30/30 with nothing non-green.

**Reviewed.** `cr@f9d96df` — CodeRabbit's summary comment updated 22:16:08Z, after the 22:13:03Z
request, naming this head, carrying no rate-limit or service marker. Its only finding section is
`🔇 Additional comments (2)`, which the contract excludes as informational. Zero review objects exist at
this head, so the `CodeRabbit — Review completed` status was treated as a corroborator only, never as
the green.

**Tried and evaluated as a user.** The executable half — the guard — was exercised directly, above. The
prose half has **no runtime surface**: it is consumed by a future run reading the Cadence section, so
per *Autonomy* I state that explicitly rather than implying I exercised it. What I did instead was read
the section end-to-end as that reader and fix three defects self-review found in my own text before
committing: an absence-based claim that the Codex store "records zero dropped dispatches" (it has no
skip surface at all, so absence there proves nothing), an over-claim that a "240-minute run" blocked
nothing when what is actually observable is a *record* that stayed open 240 minutes, and a stray line
break left mid-sentence.

Issue #2741, opened 2026-08-09
Image-payload mask and credential table still disagree on data URLs carrying an escaped newline


devantler:
> 🤖 Generated by the Agentic Engineer

## Evidence

The credential table decides whether an image payload is excluded by **parsing** the record and testing the parsed value. The locator/concentration mask decides the same thing **textually**, on the unparsed line. Two spellings still make the two answers differ, measured against the current head of #2738:

| value (as written in the record) | table excludes it | mask masks it |
|---|---|---|
| `data:image/p\ng;base64,…` (escaped newline inside the media type) | yes | **no** |
| `data:image/png;base64,…\n` (escaped newline at the very end) | yes | **no** |

Both are legal JSON, so the record parses and the per-record parse gate correctly lets the textual expressions run; the expressions then decline a value the table dropped.

## Impact

Affects anyone reading the credential-provenance output. In both cases the table does **not** count the value, while the locator still points at it — a locator with no corresponding table row, i.e. a false positive pointing at encoded image bytes. That is the same class of locator/table disagreement #2522 exists to remove, and it is the *safe* direction: no real credential is hidden. It is not urgent and does not block #2738.

## Expected behaviour

For any record that parses, the mask reaches the same verdict as the table for every value, without a new special case per JSON spelling.

## Why another regex patch is the wrong fix

Five rounds of review on #2738 each found a different JSON spelling of this same divergence (key order, letter case, escaped solidus, unicode-escaped solidus, invalid escape). The root cause is structural: a text regex approximating a predicate the other side computes by parsing has an unbounded tail, and these two are the tail continuing. #2738 closed the *record-level* half by asking the parser directly. Closing the *value-level* half needs the same move — deriving the excluded spans from the parse rather than re-deriving them textually — while preserving the properties the raw scans depend on: one process per file, line count and line numbering intact, and no second decode pass.

## Acceptance criteria

1. Both rows above agree, pinned as regression cases in the existing blob-payload fixture.
2. No new divergence in the ten cases already pinned there.
3. Line count, `grep -n` numbering and byte content of non-excluded records are unchanged.
4. Still exactly one credential-table decode pass (the batching contract test already asserts this).
5. A stated rule for what happens to a value whose exclusion cannot be determined, and evidence it fails toward *scanning* rather than toward masking.

## Size

Small-to-medium. Best designed together with #2740, which wants the same code path to stop walking every byte of multi-megabyte image lines.

devantler:
[...]

Issue #2740, opened 2026-08-09
The image-payload mask rescans multi-megabyte image lines once per session file


devantler:
> 🤖 Generated by the Agentic Engineer

## Evidence

Raised as a P2 by Codex reviewing #2738, which added `cred_mask_image_payloads` to the two raw-line
credential scans so they exclude the same complete `input_image.image_url` payloads the table does.

The mask is a `sed` with a repeated alternation group, and it runs inside the per-file loop of both
the concentration pass and the provenance locator — so it walks every byte of every embedded image
payload, twice per file. Codex measured **~2.2 seconds** for a single synthetic 5 MB matching image
line on GNU tools. Codex transcripts containing image outputs are exactly where such lines occur, and
the script's own comments describe those payloads as very large.

This is measured on a synthetic line, not on the live corpus — sizing it against real transcripts is
the first step below.

## Impact

`agent-telemetry.sh` is run by the Agent Improver on a 1-day and a 7-day corpus. A window containing
several image-bearing sessions pays this cost serially, twice over, on a script whose runtime was
already dominated once before by per-session process startup (which is why the credential table
batches its decode into a single `jq` invocation and the code carries a "DECODE EXACTLY ONCE" rule).

No correctness impact — this is throughput only.

## Expected behaviour

The raw scans exclude image payloads without walking every byte of them once per surface.

## Smallest useful first step

Measure before proposing: time the two scans over a real window with and without the mask, and record
per-file cost against payload size. Only then choose between the candidates — a cheap pre-filter so
the expensive expression runs only on lines that actually contain `input_image`; collapsing the two
raw passes so the mask is applied once rather than per surface; or a non-backtracking formulation of
the payload class.

## Acceptance criteria

- Before/after numbers in the PR body, against a measured baseline on a real corpus (not a synthetic
  line alone).
- No change to which records are masked: the nine cases pinned by #2738's contract tests still pass,
  including the malformed-escape record that must NOT be masked.
- The single-`jq`-invocation assertion for the credential table still holds.

Follow-up to #2738. Rough size: S–M.

devantler:
> 🤖 Generated by the Agentic Engineer

Measurement from #2738 that bears on this, including the part that did **not** hold up.

[...]

PR #2745, opened 2026-08-09
fix(telemetry): let the parser decide which image payloads the credential mask may blank

### LINKED ISSUE #2742: Image-payload mask trusts the first duplicate JSON key; the parser trusts the last

## Evidence

JSON permits a repeated key, and `jq`'s `fromjson` keeps the **last** occurrence. The credential table therefore reads this record's parent as `input_file`:

```
{"type":"input_image","type":"input_file","image_url":"data:image/png;base64,…"}
```

so it does **not** exclude the value and **counts** any credential-shaped substring in it. The raw-line mask matches the **first** `"type":"input_image"` textually and blanks the payload anyway.

Verified against the shipped function at `e70872e3` (not a reimplementation), with `jq` confirming the effective type is `input_file`:

| | table excludes | mask masks | result |
|---|---|---|---|
| duplicate `type`, later wins | no | **yes** | table row with **no** locator |

## Impact

This is the **unsafe** direction: the credential table reports a row while concentration reads zero and no locator points anywhere, so whoever triages it is told something was found and given nothing to look at. Distinct from #2741, whose two cases both err toward showing a pointer for a value the table dropped.

**Pre-existing** — it is a property of matching a parse-derived predicate textually, not something introduced by #2738, which strictly reduced divergence.

## Expected behaviour

For any record the table parses, the mask reaches the table's verdict — including when key precedence decides it.

## Acceptance criteria

1. The record above keeps its locator, pinned as a regression case in the blob-payload fixture.
2. A record with duplicate keys where the **later** value is `input_image` is masked, so the fix is not merely "decline anything with a repeated key" applied blindly — or, if declining is the chosen rule, that case is asserted as declined and the reasoning recorded.
3. The twelve cases already pinned there do not change.
4. Whatever the rule, an undecidable record fails toward **scanning**, never toward masking.

## Note on approach

Three review rounds on #2738 produced findings of exactly this family, each a different way a text pattern can disagree with a parser. #2741 carries the same conclusion: the durable fix derives the excluded spans from the parse rather than re-deriving them textually, and these two issues should be designed together with the perf constraint in #2740.

devantler:

## Why

The credential report could tell whoever triages it that something was found and then point at nothing. A transcript record can name one content type in its raw text while the parser reads a different one, and the two halves of the report disagreed about which is real — so a genuine credential got hidden from the locator while still being counted in the table. That is the wrong direction for a leak detector: silence where evidence should be.

## What

The parser now decides which image payloads may be hidden from the raw scan; the text-matching step only finds where they are. Records the two cannot agree on are scanned rather than hidden, so an undecidable case errs toward showing evidence. Three new regression records pin both directions, including one that only the narrower of the two new checks catches.

No behaviour change for any record already handled correctly — the twelve previously pinned cases are unchanged.

Fixes #2742
Part of #2522

### DIFF
--- a/.claude/scripts/agent-telemetry.sh
+++ b/.claude/scripts/agent-telemetry.sh
@@ -1156,9 +1156,11 @@ strip_ansi() {
 # the line. Both are conservative: a false veto merely scans a record the table
 # excluded, which is the safe direction. The vetoes are anchored on `type` keys
 # and key-colon pairs, so the `/` a payload is allowed to carry cannot
-# trigger them. What remains uncovered is a record combining an escaped key with
-# a duplicate key so the totals coincide; it is recorded rather than closed,
-# because closing it costs a whole-line rewrite per record.
+# trigger them. `\u` is the only escape that can spell a letter, so no other
+# escape opens this route. The residual is a key whose own text defeats the
+# veto's `[^"]*` span — one containing an escaped quote as well as a `\u` —
+# which is recorded rather than closed, because closing it means re-deriving
+# every excluded span from the parse and rewriting the line per record.
 #
 # The counting runs inside the SAME `jq -R` pass that already asks the parse
 # question, so this adds no process and no second read of the line.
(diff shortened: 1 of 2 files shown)


devantler:

@coderabbitai review

devantler:

@codex review




CodeRabbit refused this head with `Review rate limited`, posted after this round's request marker and with no restarting artifact between them — so the refusal belongs to this round and advancing is correct rather than inherited from another PR. `cr:no-gate@cd620fcd` recorded.

devantler:

## Readiness at `cd620fcdfd`

**Programmatically tested.** Suite 555 passed / 0 failed. RED→GREEN: before the fix, 552/1 with exactly the line-14 assertion failing; after, 555/0. Ablation on the respelling veto removes it and fails exactly the one assertion written for it (554/1) — the two gate clauses are each proven load-bearing by a case only that clause catches. Pre-existing shellcheck warnings unchanged; no new ones in the changed range.

**Reviewed.** Codex green at `cd620fcdfd`, matching head. Zero `## Review finding` comments from the connector at this head (its green and a comment-form finding can coexist, so both surfaces were checked), zero unresolved threads, zero CodeRabbit body findings at head. CodeRabbit was asked first and refused this head with `Review rate limited` — `cr:no-gate@cd620fcd`.

**Tried and evaluated as a user.** Ran the miner at this head over the live corpus: rc=0, 58,914 bytes, 373 records across 48 sessions, credential table populated by shape with the `[blob-embedded]` classification intact. Read the rendered operator note — the earlier run at a previous head is what caught its explanation going stale, so it was re-read here rather than assumed.

Also verified `git log origin/main..HEAD` carries no closing keyword, so the squash cannot auto-close `#2522` — that parent stays open for the remaining divergence work in #2741 and #2740.

PR #2790, opened 2026-08-12
fix(scripts): stop pipefail SIGPIPE inverting test-harness assertions

devantler:

### Motivation

Our shell test harnesses assert conditions by piping a value into `grep -q`. Under `pipefail` that
construct can return the wrong answer: `grep -q` stops reading at the first match, the writer is then
killed by the broken pipe, and that write error is reported as the pipeline's result.

The consequence splits in two, and the second one is the reason this matters:

- Checks written as *"this must appear"* turn into **noisy false failures**.
- Checks written as *"this must NOT appear"* silently report **success** — guard assertions that
  quietly stopped guarding anything.

The same construct is also in `agent-telemetry.sh`, which is production code rather than a test.

Scope, stated precisely: this depends on the local `grep`, so it bites the macOS agent host — where
these scripts actually run every hour — and leaves the Linux CI runners unaffected. **CI was not
red.** The repo already had a fix for this at one site, with a regression test, that was never
propagated to the other 73.

### What this changes

Every affected assertion now matches the value directly instead of piping it, so no reader can kill
a writer. The change is mechanical and one-for-one; the suites go from intermittently failing to
consistently clean, with both directions of the fault covered by controls.

### DIFF
--- a/.claude/scripts/agent-telemetry.sh
+++ b/.claude/scripts/agent-telemetry.sh
@@ -1949,8 +1949,8 @@ if want dispatch; then
       # Prose that merely quotes it is far longer, which is what keeps this
       # tool's own evidence out of its own count.
       if [ "${#lastt}" -le 200 ] \
-         && printf '%s' "$lastt" | grep -qiE "$DH_RE" \
-         && ! printf '%s' "$lastt" | grep -qiE "$DH_NOT_RE" \
+         && grep -qiE "$DH_RE" <<<"$lastt" \
+         && ! grep -qiE "$DH_NOT_RE" <<<"$lastt" \
          && { [ "$sr" = "stop_sequence" ] || [ -z "$sr" ]; }; then
         ended_on_refusal=1
       fi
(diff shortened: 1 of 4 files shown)


devantler:

@coderabbitai review

devantler:

CodeRabbit refused this head with a rate limit, so advancing to the next lane.

@codex review

devantler:

## User evaluation — exercised on the host this actually affects

The defect is macOS-only (BSD grep SIGPIPEs the writer; GNU grep reads the small stream in one block
and never does), so a green Ubuntu CI proves the change is harmless, not that it works. Ran all three
suites on the real host — bash 3.2.57 arm64-darwin, invoked as `bash <script>` so no shell shim is in
play — at the merge base `1b598b9` and at this head `d07df9ee`:

| suite | base `1b598b9` | head `d07df9ee` |
|---|---|---|
| `board-add.test.sh` | **73 passed, 7 failed** | **80 passed, 0 failed** |
| `worktree-cleanup-all.test.sh` | **8 passed, 3 failed** | **11 passed, 0 failed** |
| `worktree-cleanup.test.sh` | 30 passed, 0 failed | 30 passed, 0 failed |

Ten assertions that were reporting a **successful match as a failure** now report the truth, and the
totals are conserved in both suites (73+7=80, 8+3=11) — so the failures were converted, not skipped
or dropped.

**Stated plainly, because it is the honest reading:** the third suite is unchanged. Its 40 touched
lines carried the same unsafe idiom but none of its assertions happened to be inverting on this host,
so that part of the diff is a consistency fix whose evidence is "still 30/30, no regression" rather
than a demonstrated repair. I would rather record that than let a uniform-looking table imply three
suites were broken.

As a user of these harnesses each run, the difference is the one that matters: a red suite that had
to be second-guessed against the shell now agrees with what the code does.

PR #2791, opened 2026-08-12
fix(scripts): sweep the worktrees agents actually create

devantler:

### Motivation

The scheduled worktree cleanup only looked one level down, so it never saw a worktree created
**inside** another session's worktree — and that is where agent work worktrees are required to live,
because the write-boundary guard refuses edits anywhere else. In effect the sweep was blind to
essentially every working copy an agent creates.

Six such worktrees were sitting unswept on the host, the oldest four days old. Leaked worktrees are
not only wasted disk: a branch held by one is permanently exempt from branch cleanup too, so the
blind spot quietly disables a second piece of housekeeping. A previous leak of this kind filled the
disk and stopped new sessions from starting.

### What this changes

The sweep now finds nested worktrees and evaluates them under the existing safety rules. Verified on
the real repository: the six previously-invisible worktrees are now assessed and **none** was
removed — every one was correctly held back by an existing safety rule. So this restores visibility
without making the sweep more aggressive.

### DIFF
--- a/.claude/scripts/worktree-cleanup.sh
+++ b/.claude/scripts/worktree-cleanup.sh
@@ -161,7 +161,12 @@ record() { # path branch sha evidence outcome
   return 0
 }
 
-keep() { kept=$((kept+1)); printf 'KEEP   %-52s %s\n' "$(basename "$1")" "$2"; }
+# Label a candidate by its path RELATIVE to WT_ROOT, not its basename: nested worktrees
+# (<session>/.claude/worktrees/<name>) are swept too, and two of them under different parents
+# share a basename, so a basename-only label cannot say which one a line is about. Identical
+# to the basename for a direct child, which is every pre-existing case.
+wt_label() { local l=${1#"$WT_ROOT"/}; printf '%s' "${l:-$(basename "$1")}"; }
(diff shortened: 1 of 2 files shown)


devantler:

@coderabbitai review

devantler:

Readiness check at `fc4cf4d93b1822d0de931ff52449084adb38c982`.

Exercised against the **real** host rather than a fixture, because the whole defect is about which
directories the sweep can see:

| | Worktrees assessed | Nested ones seen | Removed |
|---|---|---|---|
| Today's script | 135 | 0 | 0 |
| This change | 142 | 7 | 0 |

The seven newly-visible worktrees are exactly the ones nested inside another session's worktree —
where agent working copies are required to live. Every one of them was **kept**, each for a reason
the existing safety rules already supply (unpushed commits, uncommitted work, too recent). Nothing
was removed before or after.

That is the property worth stating plainly: this makes the sweep see more, not delete more. The risk
in a change like this would be a newly-visible worktree being reaped on sight; the run shows the
opposite, with the fail-closed rules doing their job the moment visibility is restored.

Suite is 31 passed / 0 failed at this head, including a regression test that fails on the old script.
Reviewed clean by CodeRabbit at this head, all checks green.

Issue #2779, opened 2026-08-11
Dependabot's submodule queue has been stalled since 2026-08-05, freezing every submodule pin including the agent-definition source


devantler:
> 🤖 Generated by the Agent Improver

## Evidence

Measured 2026-08-11T22:21Z against `devantler-tech/monorepo`.

Dependabot's `gitsubmodule` ecosystem has stopped delivering entirely:

| | value |
|---|---|
| Newest submodule bump PR **created** | 2026-08-06T20:14:14Z (**5d 2h** ago) |
| Newest submodule bump PR **merged** | 2026-08-05T20:15:10Z (**6d 2h** ago) |
| Submodule bump PRs merged in the 14 days before that | **26** |
| Open submodule bump PRs right now | **5** — #2683, #2700, #2701, #2702, #2703 |

Five is Dependabot's default `open-pull-requests-limit`, and `.github/dependabot.yaml` does not
override it. The ecosystem is saturated, so the daily job creates nothing new.

**The queue cannot drain on its own.** All five are `BLOCKED`:

- **#2700, #2701, #2702, #2703** — `gh run list --branch <head>` returns **no runs at all**. Their
  entire status rollup is one `CodeQL` check at `NEUTRAL`; the required CI workflows and
  `EnableAutoMerge` never fired, so required checks can never report and auto-merge was never armed.
- **#2683** — did run CI (31 checks) and has auto-merge armed, but `drift-check-active-projects`
  (`CI - Required Checks`) is failing.

#2683 was created 2026-08-05 and got its checks; #2700–#2703 were created 2026-08-06 and got none —
so workflow dispatch on Dependabot submodule branches changed between those two daily runs.

**Control that rules out "Dependabot is broken".** The same daily job window created an npm PR
(#2775) at **2026-08-11T20:15:27Z** — one day ago. Dependabot is alive and creating PRs; the npm
ecosystem sits at 4 open, under its own limit of 5. One invocation, two ecosystems: the saturated one
produced nothing, the unsaturated one produced a PR.

## Impact

**Every submodule in `.gitmodules` is now frozen at its 2026-08-05/06 pin, indefinitely** — this is not
a backlog that clears, because four of the five blockers are structurally unmergeable.

The sharpest casualty is `libraries/agent-plugins`, the **source of this deployment's own agent
definitions**. It is pinned at `bb17f468` (2026-08-03) while upstream `main` is `ebd6a62` (2026-08-10),
[...]

PR #2795, opened 2026-08-12
chore(deps): bump agent-plugins pin to 0ecb7b23

devantler:

### Motivation

The automation that keeps our submodule pins current has proposed nothing since 2026-08-06 — it is jammed at its open-PR limit by PRs that cannot drain. While it has been stuck, the pin for the shared agent-plugins library has sat five commits behind, so every scheduled run has been loading agent definitions that are a day and a half out of date, including a security fix to the surveyor that landed today.

### What

Moves that one pointer forward to the current, CI-green tip of the library. No other change.

This is done on our own branch rather than by touching the automation's PRs, which stay hands-off — the jam itself is tracked separately and is not fixed here.

Part of #2779


devantler:

@coderabbitai review



Single gitlink bump, `bb17f468` → `0ecb7b23`, no other file touched.

The target is the current tip of `agent-plugins` `main` and its own CI is green there (both `🧪 CI` and `Push on main` completed successfully at that commit). The five intervening commits are all merged upstream PRs. The full 40-character sha was resolved from the API rather than expanded from an abbreviation.

CI here is green and the PR is CLEAN.

devantler:

@codex review




CodeRabbit was requested at this head at 15:39:37Z, acknowledged, and then refused with `Review rate limited` at 15:40:03Z — a refusal this round's own request produced, so the lane is demonstrably not serving here and this advances rather than re-spending it.

Worth noting for whoever reads this later: this head first carried `Review skipped: ignored keyword in the PR title`, so the org config excludes `chore(deps):`-titled PRs from automatic review. The explicit request did override that, which is why the rate limit is the operative refusal rather than the title filter. The title is not being changed to dodge the filter — `chore(deps):` is the correct Conventional Commit type for a pin bump and it drives the changelog.

Scope for review: one gitlink, `bb17f468` → `0ecb7b23`, nothing else. CI green, PR CLEAN.

devantler:

**Readiness record for `789ac1145e`.**

**Programmatically tested** — CI green at this head (12 passed, 25 skipped), `mergeStateStatus` CLEAN, 0 unresolved threads, 0 non-thread review findings.

**Reviewed** — Codex green at this head: *Didn't find any major issues*, `Reviewed commit: 789ac1145e`, matching `headRefOid`. Swept for comment-form `## Review finding` sections as well; none. CodeRabbit did not deliver here: this head first read `Review skipped: ignored keyword in the PR title` (org config excludes `chore(deps):` titles), the explicit request overrode that and was acknowledged, then it refused with `Review rate limited` — a refusal this round's own request produced, so the lane was demonstrably not serving and the loop advanced rather than re-spending it.

**Tried and evaluated as a user** — this is a pointer bump, but it is not without an observable effect, so I did not fall back on the no-runtime-surface carve-out. I resolved the newly-pinned tree and confirmed it is what a run would actually load: all three agent definitions are present, including the `agentic-engineer` entrypoint the contract requires be verified at the merged upstream tip. I then diffed the surveyor definition across the two pins — its read-only boundary text differs between the old pin and the new one, i.e. the security fix that motivated this bump is genuinely present in the tree being pinned and genuinely absent from the one being replaced. So the bump is not a silent no-op.

One thing worth stating plainly: the full 40-character sha was resolved from the API. My first attempt staged a sha I had extended from a 10-character abbreviation by hand, which was a fabricated value; I caught it before committing and re-derived it properly.

Issue #2787, opened 2026-08-12
fix(scripts): a matching grep -q can fail an assertion via SIGPIPE under pipefail


devantler:
> 🤖 Generated by the Agentic Engineer

## Evidence

Two contract tests are red on `main` right now (`1b598b9`), and at least one of them is red for a
reason that has nothing to do with what it asserts.

**Proven — `review-provider-loop-contract.test.sh` fails deterministically (5/5 runs):**

```
review-provider loop contract: FAIL — fixture clean-pass comment no longer names the same head
```

The message is misleading: the fixture *does* name that head. Instrumenting the assertion shows the
pipeline returning **141** — `grep -Fq` matches, exits immediately, and the still-writing `printf` is
killed by SIGPIPE. Under the script's `set -o pipefail`, the pipeline then reports **printf's** 141
rather than grep's 0, so a *successful* match reads as a failed assertion:

```
DBG grep_status=141        # printf '%s' "$green_body" | grep -Fq 948bb06f73
```

**Observed, cause NOT yet attributed — `board-add.test.sh` fails nondeterministically:** 6, 7 and 6
failures across three consecutive runs of the unchanged file. It contains the same construct 16
times, so it is the obvious suspect, but I have not instrumented it and am not claiming it.

**Scope:** 83 occurrences of `… | grep -<flags>q …` across 7 scripts that set `pipefail`:
`agent-telemetry.sh` (6), `agent-telemetry.test.sh` (2), `board-add.test.sh` (16),
`branch-cleanup.sh` (1), `review-provider-loop-contract.test.sh` (4),
`worktree-cleanup-all.test.sh` (14), `worktree-cleanup.test.sh` (40). Most sit inside `if`
conditions, where `set -e` is suspended but `pipefail` still applies — so the failure mode there is a
silently-taken `else` branch rather than an abort.

## Impact

This is a **guard-integrity** problem, not a cosmetic one. `branch-cleanup.sh` and
`agent-telemetry.sh` are production scripts, and `branch-cleanup.sh` deletes remote branches on
positive evidence — a condition that evaluates as "no match" when the match actually succeeded is
exactly the wrong direction for an evidence gate to fail in. It also trains every future run to read
a red contract test as noise, which is how a real regression gets waved through.

The race is size- and load-dependent: small inputs usually let `printf` finish first, which is why
these have passed historically and why the count varies between runs.

## Expected

An assertion's result depends on whether the pattern matched, never on whether the writer lost a race
[...]

Issue #2688, opened 2026-08-05
79 pipe-into-quiet-grep sites remain under pipefail, so assertions still report a match as a non-match


devantler:
> 🤖 Generated by the Agentic Engineer

## Evidence

Under `set -o pipefail`, `writer | grep -q needle` reports the pipeline as **failed when the needle is found**: `grep -q` exits at the first match, the still-writing producer takes `SIGPIPE` (141), and `pipefail` surfaces that as the pipeline status.

Two instances of this construct were fixed on 2026-08-05 and both were severe:

- **#2673** — `agent-telemetry.test.sh` ran 132 assertion pipelines this way. Repairing them moved the suite from **190 passed / 310 failed / 296 broken-pipe** to **507 / 0 / 0**. Worse than the noise: every `nocheck` absence assertion passed *vacuously*, including the ones proving a credential does not survive redaction — a leak detector that could not fail.
- **#2675** — `branch-cleanup.sh` used it for the pre-delete check that keeps a branch a live worktree claimed. A branch that **was** checked out read as checked out nowhere. Correct at ≤2 worktree entries, wrong from ~5 up; this host enumerates ~124.

The remaining sites were never swept, and #2674 — where the sweep was recorded — closed when #2675 merged, so nothing tracks them.

## Scope, re-derived on `main` at 2026-08-05

| file | code-level instances |
|---|---|
| `.claude/scripts/worktree-cleanup.test.sh` | 40 |
| `.claude/scripts/board-add.test.sh` | 16 |
| `.claude/scripts/worktree-cleanup-all.test.sh` | 14 |
| `.claude/scripts/agent-telemetry.sh` | 6 |
| `.claude/scripts/review-provider-loop-contract.test.sh` | 3 |
| **total** | **79** |

All five set `pipefail`.

## Two corrections to the earlier note

- The count was previously recorded as 80–84. The detector needs a leading `[^|]` (or it matches the second pipe of `||`, where `grep` reads a file and there is no writer to kill) **and** must skip comment lines, or it counts prose that merely quotes the construct.
- `worktree-claim.test.sh` was recorded as unaffected because it "sets no pipefail". It **does** set `set -Eeuo pipefail` on line 5. It is nonetheless clean: both of its matches are comments explaining why the construct is avoided there. The exemption was right, the reason was wrong.

Detector:

```
[...]

Issue #2363, opened 2026-07-21
Consume the automated-ai-engineer plugin instead of vendoring its definition


devantler:
> 🤖 Generated by the Agentic Engineer

**Part of** devantler-tech/agent-plugins#51 — the third and only outstanding phase of the
"extract the Automated AI Engineer core into a plugin" epic.

## Evidence

The plugin exists and is complete; the monorepo does not use it. Verified on main, 2026-07-21:

- **Phase 1 (design ADR) — delivered:** `docs/adr/0002-automated-ai-engineer-plugin-boundary.md`.
- **Phase 2 (extract) — delivered:** `plugins/automated-ai-engineer/` v1.0.0 ships 3 agents
  (`automated-ai-engineer`, `portfolio-surveyor`, `agent-improver`) and 4 skills
  (`portfolio-maintenance`, `product-engineering`, `self-improvement`, `agent-improvement`).
- **Phase 3 (consume) — NOT started:** this monorepo installs nothing from that plugin. It still
  carries its own `.claude/agents/{daily-maintainer,portfolio-surveyor,agent-improver}.md`, its own
  copies of all four skills, and a 2281-line `AGENTS.md` that mixes the generic role with this
  deployment's configuration.

## Problem, and who it hurts

Two full copies of one definition now exist, and the generic half is the half that changes weekly.
Every improvement to the role has to be made twice or it drifts — and drift here is not cosmetic:
these files are what steer three live agent instances. The epic named this risk itself ("a bad split
leaves half-generic text in both places that drifts"); the split has now happened, so the risk is
live rather than hypothetical.

It also leaves the plugin unproven. Nothing currently consumes `automated-ai-engineer`, so its
parameterization contract — the consuming repo's `AGENTS.md` supplies the portfolio map, trust gate,
cadence and channels — has never been exercised by a real consumer. The monorepo is the intended
flagship proof of that contract.

## Success signal

The generic role is edited in exactly one place. Concretely: the sections this issue moves are
present in the plugin and absent from the monorepo's own files, and a run of the daily engineer
completes normally against the plugin-provided definition.

## Smallest useful first slice

**Do not shrink `AGENTS.md` in one pass.** Start by consuming the plugin *alongside* the local
definition and proving they agree, before deleting anything:

1. Install the marketplace/plugin in this repo's Claude configuration.
2. Pick the **one** least deployment-coupled artifact — the read-only `portfolio-surveyor` agent —
[...]

PR #2820, opened 2026-08-13
fix(agents): identify CodeRabbit reviews behind the agent-hint prefix

### LINKED ISSUE #2819: CodeRabbit review bodies now start with an agent-hint comment, so positive identification false-negatives

## Evidence

CodeRabbit now prefixes every real review body with an HTML comment, so the body no longer *begins*
with the string the contract uses to identify it as a review:

```
<!-- coderabbit-cli-agent-hint:v3
After fixes: `coderabbit review ...`. Missing? Ask user; ...
-->

**Actionable comments posted: 5**
```

Measured 2026-08-13 on four substantive review objects — monorepo#2810 at 07:18:36Z (len 30713) and
10:56:58Z (len 37894), monorepo#2723 at 2026-08-12 20:33:20Z (len 49012) and 21:59:21Z (len 62825).
All four carry the prefix; none begins with `**Actionable comments posted:`.

Every run, on every PR — this is the gate that decides whether a CodeRabbit review counts.

The test **fails closed**, which is the safe direction but not a cheap one: a genuine current-head
review reads as "no review", so the run walks down the lane order and spends **weekly-limited Codex**
and **monthly-limited Bugbot** on a head CodeRabbit has already reviewed. That is precisely the
inversion the cheapest-lane-first order exists to prevent, and the contract already records the same
failure shape for the verdict-reply case. A pentad-clear PR can also park while reading
`green_review=none` over a real green.

## Expected behaviour

Positive identification should tolerate leading HTML comments: strip any leading `<!-- … -->` blocks
(and surrounding whitespace), then apply the existing `**Actionable comments posted:` test. The
identification requirement itself must **not** be weakened to a bare `commit_id` match — the
empty-container measurement that motivated it still stands.

## Acceptance criteria

- [ ] Prefix-tolerant matching stated in `AGENTS.md` (2 sites: the lane table and the
      *CodeRabbit success* paragraph).
- [ ] Same in `.claude/agents/portfolio-surveyor.md` (2 sites) and
      `.claude/plugin-consumption/agentic-engineering-surveyor-diff.md` (1 site).
- [ ] A guard fixture pinning a body that carries the hint prefix, asserting it identifies as a
      review — and a negative control that an empty container still does not.
- [ ] The `commit_id`-alone weakening remains rejected.

Rough size: S — a matching-rule change across five prose sites plus one guard.

devantler:

### Why

Our rule for recognising a finished CodeRabbit review stopped matching reality: CodeRabbit now puts a
hidden hint block at the top of every real review, and our check required the review to *start* with a
particular marker. So genuine reviews were being read as "no review at all".

The cost is real money and time. Every run then re-asked the free reviewer, then spent the
weekly-limited one and the monthly-limited one on a pull request that had already been reviewed —
exactly the waste our cheapest-reviewer-first ordering exists to prevent — while a finished PR sat
waiting.

### What

Ignore the hint block before checking for the marker, in all five places the rule is written down.
Nothing is loosened: the check still has to match at the start, and the existing protection against
counting an empty placeholder as a review is untouched.

The guard uses a real captured review as its test fixture and keeps the old rule alongside it, so if
the fixture ever stops reproducing the problem the test fails instead of quietly passing.

Fixes #2819


devantler:



CI is green at `80b0999b` (23 success / 14 skipped), including the two contract jobs this change
touches. Self-review done before requesting.

@coderabbitai review

devantler:

**Readiness record for `80b0999b`** (kept out of the body, which stays PM-level).

**Programmatically tested — RED/GREEN, with the ablation verified.**
- RED: the guard failed on the unfixed prose (`constitution's CodeRabbit identification is not
  prefix-tolerant`).
- GREEN: all assertions pass after the five-site fix, and CI ran **Test review-provider loop
  contract** and **Test portfolio surveyor contract** green at this head.
- The ablation was *checked to fire*, not assumed: replacing the fixture with a hint-free body makes
  the guard fail as vacuous (`the captured fixture no longer reproduces #2819`). The fixture is real
  captured output, and the pre-fix rule is kept live so it must keep failing on it.
- Negative controls: empty container, hint-only body, chat acknowledgement, and a mid-body marker
  mention are all still rejected — the match stays anchored rather than becoming a substring search.

**Tried and evaluated as a user — end-to-end on a real review, not the fixture.** Applied both rules
to an independent substantive CodeRabbit review (#2810, body length 37894):

```
pre-fix rule  -> none
fixed rule    -> IDENTIFIED
```

That is the whole defect in one line: a genuine review reading as `green_review=none`, which is what
sent runs down into the weekly- and monthly-limited lanes on an already-reviewed head.

**One local-environment note, deliberately not "fixed" here.** This guard cannot run to completion on
macOS: an unrelated pre-existing assertion uses `printf … | grep -q` under `set -o pipefail`, and
`grep -q` exiting early gives `printf` SIGPIPE (`rc=141`), with real `/usr/bin/grep` as well as the
host's shim. It fails identically on clean `main`, and CI runs the same job green on Linux, so it is
a host-only artifact rather than breakage — I verified locally by disabling `pipefail` alone and left
the assertion untouched. Filed separately rather than mixed into this PR.

devantler:

Correction to the note above: I wrote "filed separately", but this class is already tracked — it is
[#2688](https://github.com/devantler-tech/monorepo/issues/2688) (79 remaining pipe-into-quiet-grep
sites under `pipefail`), with [#2787](https://github.com/devantler-tech/monorepo/issues/2787) covering
the same failure mode. Filing a third would have been duplicate noise, so I added this specific site
and its reproduction to #2688 instead. The assertion itself stays untouched in this PR.

devantler:



CodeRabbit acknowledged the invocation at 14:28:38Z and refused at 14:28:41Z, after this round's
request marker, so the refusal belongs to this round. It reports all free OSS reviews used with a
**69-minute** reset — longer than this run, so waiting would park a finished change rather than
shorten anything. Recording `cr:no-gate@80b0999b` and advancing to Codex.

devantler:



CI green at `80b0999b`, 0 threads, CLEAN. CodeRabbit is quota-exhausted at this head (69-minute
reset). Worth a close look at the strip helper's loop termination and at whether the negative
controls really pin an anchored match rather than a substring search.

@codex review

devantler:

**Resolution record for the three Codex findings** (comment-form, so there are no threads to resolve).
All three were valid and are fixed at `303413f9`.

**P2 — multiline anchoring control.** Correct: my control displaced the marker only *horizontally*, so
a line-anchored implementation would have passed it while accepting a body whose marker merely starts
a later line. Added a multiline control. **Ablated it:** switching the matcher to also accept
`*$'\n''**Actionable comments posted:'*` trips exactly that control and nothing else.

**P2 — pin every operational occurrence.** Correct, and this was the most valuable of the three:
`assert_prose` flattens the whole file, so the lane table or the surveyor's primary directive could
regress to the unstripped predicate while a later explanatory paragraph kept the phrase green. Each of
the five sites now has its own needle, verified to match **exactly once**. **Ablated it:** reverting
only the AGENTS.md lane table trips its own assertion by name.

**P2 — unterminated-comment path.** The hang risk was real, and I fixed it at the root rather than
around it: the strip loop now carries an **iteration bound**, so removing the break can no longer
produce an infinite repeat — `timeout` is not installed on macOS, so an external bound was not
available.

One correction to the finding's premise, stated because it changes what the added control proves:
**deleting the break does not make the new control fail.** I checked. With the bound in place the body
comes back unchanged, still fails the anchored match, and the control passes either way. So it pins
the *rejection* of a malformed body — worth having — and **not** the break; the bound is what removes
the hang. The comment in the file now says exactly that rather than implying an ablation it does not
perform. An over-consuming strip (`${body#<!--}`) does trip the suite, so the strip logic itself is
pinned.

devantler:



Round 1's three findings are fixed at `303413f9` with a resolution record posted above. CI is green
(24 success / 14 skipped), 0 unresolved threads.

New head, new round — the `cr:no-gate@80b0999b` recorded earlier does not pre-empt this round's first
request.

@coderabbitai review

devantler:



CodeRabbit accepted the invocation at 14:55:35Z (`I will review pull request #2820 at 303413f9`) and
the head's status went `Review rate limited` three seconds later, after this round's request marker —
so the refusal belongs to this round. The account-wide free-OSS window reported at 14:28Z has not yet
reset. Recording `cr:no-gate@303413f9` and advancing to Codex.

devantler:



Round 2 at `303413f9`: your three P2s are fixed with a resolution record above. CI green
(24 success / 14 skipped), 0 threads. CodeRabbit is quota-exhausted at this head.

Note one correction recorded there: deleting the `break` does **not** make the new malformed-body
control fail — the iteration bound stops the loop and the body still fails the anchored match — so
that control pins the rejection, not the break.

@codex review

devantler:

**Resolution record for the two Codex findings at `303413f9` → fixed/refuted at `80a5d6de`.**

**P2 — `review-provider-loop-contract` path filter (`review-provider-loop-contract.test.sh:173`): VALID, FIXED.**
The captured review bodies are inputs to that test, so a fixture-only edit could invalidate a
regression case while the job stayed skipped and CI stayed green. `.claude/scripts/fixtures/**` now
joins the filter, and the contract test asserts its presence so the gap cannot reopen silently.
Verified in isolation, because this suite fails earlier on macOS for an unrelated reason (the
`printf | grep -Fq` + `pipefail` SIGPIPE artifact tracked in #2688 — it fails identically on clean
`main`, and Linux CI runs it green): the assertion's exact `grep` passes against the workflow and
fails against a copy with the line removed.

**P1 — "land the parser change in the canonical plugin first": REFUTED, on measured evidence.**
The premise is that this is portable behaviour that belongs upstream in `agent-plugins`. It is not
portable in this deployment's architecture, for two reasons:

1. **The plugin entrypoint is deliberately provider-neutral.** Read at the pinned revision
   (`a0e6ad85`), `agentic-engineer.agent.md` is 208 lines and contains **zero** occurrences of
   "CodeRabbit" — and zero of any other review provider. It speaks only of "a green review at the
   current head". Moving CodeRabbit's review-body parsing into it would put one vendor's output
   format into the tool-neutral portable role, which is the portability principle this repository's
   `AGENTS.md` states for the shared libraries ("industry-standard and tool-neutral").
2. **The surveyor overlay is the declared home for exactly this.** `AGENTS.md` records
   `.claude/agents/portfolio-surveyor.md` as a compatibility overlay that "preserves the
   deployment-hardened procedure and **output grammar** that the generic plugin does not carry yet",
   removable only once the parity checklist passes. This change updates that checklist
   (`agentic-engineering-surveyor-diff.md`) in the same commit, which is the sanctioned path rather
   than a bypass of it.

On the gitlink: bumping `libraries/agent-plugins` would not deliver this change either, since the
behaviour does not exist upstream — the bump is ordinary dependency currency, tracked separately.

If the generic kernel here ("identify a review artifact positively rather than trusting a
`commit_id` match") is later extracted, #2363 is the migration-inventory issue that owns it.

devantler:



Round 2. Both Codex findings from `303413f9` are resolved (one fixed, one refuted with measured
evidence — see the resolution record above). CI is green and the PR is CLEAN at this head.

@coderabbitai review

devantler:

**Resolution record for the CodeRabbit 🟠 Major at `80a5d6de` → fixed at `6589627f`. Valid.**

The strip was under-specified. CodeRabbit prefixes a real review body with the agent-hint HTML
comment **followed by a blank line**, so an implementation that strips only the comment leaves the
body starting with a newline and the anchored `begins **Actionable comments posted:` match still
fails — the exact defect this PR exists to close, reintroduced one character further along.

Verified on two independent bodies rather than reasoned: the captured fixture and **this PR's own
current-head review body** both have `-->` on one line and an empty line before the marker. The
reference implementation in the contract test already stripped surrounding whitespace; only the prose
did not, so someone implementing from the prose would have built the failing version.

**Swept by grep rather than by the finding's list**, which found **five** sites, not the two named:
both constitution sites, both surveyor sites, and the parity checklist. Two of them were
**hard-wrapped** across a newline, so a line-based substitution silently skipped them — caught because
the contract test normalises whitespace before matching and failed on each in turn until all five
moved together, along with its own pinned needles.

The suite now reaches the pre-existing macOS `printf | grep -Fq` + `pipefail` SIGPIPE failure tracked
in #2688 — it fails identically on clean `main`, is a host artifact rather than a regression, and
Linux CI runs this job green.

devantler:



Round 3 at `6589627f`. The 🟠 Major from round 2 is fixed: the hint-prefix strip now also removes the
blank line that follows the HTML comment, and the same rule was propagated to all five sites that
encode it (two in the constitution, two in the surveyor, one in the parity checklist) rather than the
two the finding named. Contract suites pass; CI is green (32 success / 9 skipped) and there are no
unresolved threads.

@coderabbitai review

devantler:



**CodeRabbit did not review `6589627f` — recording `cr:no-gate@6589627f` and advancing to Codex.**

The trigger was answered with the chat-shaped reply (`[!TIP] For best results, initiate chat on the
files or code changes`) rather than a review-command invocation. The head's `CodeRabbit` commit
status has not advanced past its pre-request value, and there is no CodeRabbit review object at this
head. Per the provider loop this is a service outcome with no code issue to fix, so the round
continues at the next lane rather than re-spending the free lane.

devantler:



Round 3 at `6589627f`. CodeRabbit's 🟠 Major from round 2 is fixed — the hint-prefix strip now also
consumes the blank line that follows the HTML comment, and the rule was propagated to all five sites
that encode it rather than the two the finding named. Contract suites pass, CI is green
(32 success / 9 skipped), no unresolved threads, CLEAN against base.

@codex review

devantler:

**Readiness record at `6589627f` — all three conditions proven, promoting.**

**1. Programmatically tested.** CI is green at this head (32 success / 9 skipped, 0 failures), and
the job that actually *consumes* this change ran rather than being filtered out:
`Test review-provider loop contract | success | 15:59:02Z`. That matters here specifically, because
the change adds `.claude/scripts/fixtures/**` to that job's path filter — a green CI in which the
job had skipped would have proven nothing.

**2. Reviewed.** Codex returned a clean current-head pass at 16:31:10Z —
`Didn't find any major issues`, `**Reviewed commit:** 6589627f59`, matching `headRefOid`. CodeRabbit
was asked first per lane priority and answered with the chat-shaped reply at 16:21:36Z without
running a review (head status never advanced, no review object at head), so it was recorded
`cr:no-gate@6589627f` and the loop advanced. No Codex `## Review finding` comment exists at this
head, and there are zero unresolved threads.

**3. Tried and evaluated as a user.** I ran the shipped identification functions against a **live**
CodeRabbit review body pulled from the API just now — the 45,737-byte review of this very PR at
`80a5d6de` — rather than only the committed fixture:

| input | pre-#2819 rule | shipped rule |
|---|---|---|
| live 45,737-byte CodeRabbit review body | **NO MATCH** (defect reproduces on live data) | **MATCH** |
| empty reply container | — | NO MATCH ✓ |
| chat acknowledgement | — | NO MATCH ✓ |
| hint comment only, no marker | — | NO MATCH ✓ |
| marker present but not at start | — | NO MATCH ✓ (still anchored) |
| unterminated `<!--` before the marker | — | NO MATCH ✓ (strip stops, does not consume the body) |

So the rule recognises a real hint-prefixed review, the defect it was filed against still reproduces
on freshly-fetched data, and none of the four protections the widening could have eroded — empty
container, chat ack, anchoring, unterminated comment — actually eroded.

PR #2865, opened 2026-08-16
fix(agents): put one verb in a Bash call

devantler:

### Why

The runtime's auto-mode classifier refuses a share of our `Bash` calls outright, and the refusals are
not random: they concentrate on calls that put setup in front of the real command. Over the 7 days to
2026-08-16, across the Claude session corpus, there were **58 classifier denials in 15 distinct
sessions**. The **27 multi-statement** ones span **12 sessions**, and **25 of those 27 begin with
`cd ` (18) or a variable assignment (7)** before the verb that matters. Every denial loses the whole
call, so the run pays for the setup too and then re-derives the same command a second way.

It is worth fixing because it costs nothing to comply. The prefix is almost always unnecessary — the
Bash tool's working directory persists between calls, and `git -C`, `gh --repo` and absolute paths
already carry their own location — so the habit can go without losing any capability.

The contract already stated this for exactly one case (a compound `fetch`+`checkout`, in *Git
safety*). This makes it the general rule, where the measurement says it is actually needed.

### What

Adds a *Latency discipline* bullet telling the agent to put one verb in a `Bash` call, naming the
location-carrying alternatives, warning that the denial is **probabilistic** (which is why the habit
survives — the same spelling usually works), and stating that the recovery is to **split** the call
rather than re-spell it. A new contract test pins the six claims that make the rule actionable, wired
into CI beside the existing latency and control-character contract tests.

No guardrail is loosened: the classifier is behaving correctly, and the bullet explicitly forbids
routing around it via the permission surface.

### DIFF
--- a/.claude/scripts/one-verb-per-bash-call-contract.test.sh
+++ b/.claude/scripts/one-verb-per-bash-call-contract.test.sh
@@ -1,0 +1,95 @@
+#!/usr/bin/env bash
+#
+# Guards the *Latency discipline* bullet that stops a `Bash` call being refused by the auto-mode
+# classifier because it prefixes the real command with `cd` or a variable assignment.
+#
+# Why this needs a guard. Measured over the 7 days to 2026-08-16 across the Claude session corpus:
+# 58 auto-mode classifier denials across 15 distinct sessions. The 27 multi-statement ones span 12
+# sessions, and 25 of those 27 open with `cd ` (18) or a variable assignment (7) before the verb that
+# matters. Each denial loses the whole call, so the setup in front of the verb is discarded with it.
+#
+# The subtle part, and the reason a presence-only check is not enough: the guard is PROBABILISTIC on
+# this shape. The same `cd … ; <verb>` spelling usually succeeds, so a lane reads the occasional
+# refusal as noise and keeps the habit — one measured session led every Bash call with `cd <worktree>`
+# and paid for it repeatedly. A bullet that merely says "prefer one verb" without pinning (a) that the
+# prefix is unnecessary because the working directory persists, and (b) that the recovery is to SPLIT
+# rather than re-spell, leaves the reader with no reason to change and no working next step.
+#
+# Assertions are scoped to the bullet, not the whole file — asserting against the whole constitution
+# is a scope hole, since an unrelated section containing the phrase would satisfy the check while the
+# real sentence stayed wrong.
+
+set -euo pipefail
+
+repo_root="$(cd "$(dirname "${BASH_SOURCE[0]}")/../.." && pwd)"
+constitution="${repo_root}/AGENTS.md"
+
+fail() {
+  echo "one-verb-per-bash-call contract: FAIL — $*" >&2
+  exit 1
+}
+
+[ -r "${constitution}" ] || fail "cannot read ${constitution}"
+
+# Extract ONLY this bullet, then flatten it: the sentences wrap across source lines, so a fragment
+# spanning a line break would never match and the test would be always-red regardless of content.
+# The end anchor is the raw-control-byte bullet that follows it.
+bullet="$(
+  awk '
+    /^- \*\*Put ONE verb in a / { inb = 1 }
+    inb && /^- \*\*A raw /      { inb = 0 }
+    inb                         { print }
+  ' "${constitution}" | tr '\n' ' ' | tr -s '[:space:]' ' '
+)"
+
+[ -n "${bullet}" ] ||
+  fail "could not locate the one-verb-per-call bullet — the extraction anchor moved, so every assertion below would be vacuous"
+
+# Guard the extraction itself. If the terminating anchor disappears, awk runs to end-of-file and
+# `bullet` becomes the rest of the contract, silently restoring the scope hole while every assertion
+# still passes. The bound separates two states an order of magnitude apart (this bullet is ~250 words;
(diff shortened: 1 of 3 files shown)


devantler:



@coderabbitai review

CI is green at `e5e015ec` (41 success / 9 skipped), including the new `Test one verb per bash call
contract` job, which is confirmed to have executed rather than been filtered out. All 7 contract
assertions plus both extraction guards were ablated individually and every one fired.

Self-review before requesting found and fixed three issues in my own diff, the substantive one being
that the first commit asserted the Bash working directory "persists between calls" (as the runtime
docs state) and advised setting it once. Probing showed that is false at the workspace boundary — a
`cd` inside the session worktree persists, one outside it is silently reset — so that advice would
have had relative paths resolve in the wrong tree while still reporting success. It is replaced by
passing the location to the command, with absolute paths named as the form correct on both sides.

devantler:



@coderabbitai full review

devantler:



**Parked on a named, live-verified blocker: no review lane will deliver at `e5e015ec`.**

Everything else is clear. CI is green (41 success / 9 skipped), the new `Test one verb per bash call
contract` job is confirmed to have *executed* rather than been filtered out, there are no unresolved
threads and no non-thread findings, and the branch is `CLEAN` against base.

| lane | state at `e5e015ec` | evidence |
|---|---|---|
| CodeRabbit | **rate limited** | head status `success — Review rate limited`, read directly. Two no-gate outcomes here: the first request was answered in *chat* mode (retrieved learnings, no review, head status unchanged), the second parsed as a command and then hit the limit. The lane served another PR twice in the preceding 20 minutes and serves roughly one review per hour. |
| Codex | usage limit reached | its own comment earlier today, no retry window; needs credits enabled for code reviews |
| Bugbot | usage limit reached | `Bugbot couldn't run - usage limit reached` earlier today; an admin must raise the Cursor limit |

**Deliberately not self-reviewed into a green.** A local review round is the documented fallback when
no lane will deliver, and I am not taking it here: I have already reviewed this diff once and it was
not a formality — it found three real problems in my own work, including a false claim about the Bash
working directory that would have sent readers to a silently-wrong fix. Signing off on my own
correction of my own error, with no independent eye, is exactly the situation the external gate exists
for. CodeRabbit is free and recovers within the hour, so this is a wait, not a block.

**Next run, cheapest correct path:** re-read the `CodeRabbit` status description at whatever the head
is, request once, and expect the lane to have recovered. Do **not** spend Codex or Bugbot — both are
billing-exhausted rather than throttled, and only the maintainer can lift them.

devantler:



CodeRabbit reports a rate limit on this head, so advancing to the next lane in priority order.

@codex review

devantler:



CodeRabbit is rate limited and Codex reports its usage limit at this head, so trying the last lane.
The bare trigger follows in its own comment (Bugbot exact-matches the whole comment body).

devantler:
@cursor review

devantler (commented):

**Reviewed commit:** `e5e015ec6330e7b69df1f984d76f0031cee21a1b`

**Per-lane evidence, all measured at THIS head (not inherited from another PR):**
- **CodeRabbit** — head commit status reads `Review rate limited`. Its own retry window from the earlier 22:28:39Z refusal had expired, and the lane was spent on `monorepo#2549` at 01:10Z (it serves roughly one review per hour portfolio-wide).
- **Codex** — requested at this head 01:19:09Z; the connector replied 01:19:18Z that the account has reached its usage limit for code reviews. No retry window stated.
- **Cursor Bugbot** — requested at this head 01:19Z; check-run `Cursor Bugbot` returned `conclusion: neutral` with `output.title: Error`, and the accompanying comment reports a usage/spend limit that only an admin can raise.

All three are quota/billing states rather than verdicts on this change, so the local round applies.

### What I exercised

Ran the new contract test at this head: **7/7 assertions pass**. Then ablated the three parts most likely to be wrong rather than trusting the green:

1. **Is the `case` pattern literal or a glob?** Assertion 1 pins `loses the **whole call**`, which contains `**`. Removed just the emphasis markers (leaving the prose meaning intact) → test went **RED** with *"does not state that the entire Bash call is lost"*. So the quoted `*"$1"*` form is matching literally, as intended; a glob would have passed and made the assertion near-vacuous.
2. **Does the runaway-extraction guard actually fire?** Renamed the end anchor the way a future edit plausibly would (`- **A raw ` → `- **Any raw `) → extraction ran to end-of-file and the guard fired: *"bullet extracted as 7985 words, which is runaway-extraction size"*. That is the scope hole closed, with a diagnostic that names the cause.
3. **Restore → GREEN** after each, so neither result was an artifact of a broken file.

CI wiring checked against its siblings: the new job is declared in `changes` outputs, given a path filter covering `AGENTS.md` + its own test + `ci.yaml`, and added to both the aggregate `needs` list and the result roll-up — consistent with the existing latency and control-character contract jobs.

### Correctness note on the claim the rule rests on

The bullet's cwd warning is **right**, and it is the part I checked hardest because the runtime's own documentation states "working directory persists between calls" without qualification. The shipped prose does not repeat that unqualified claim: it states the boundary (persists inside the session worktree, silently reset outside it) and names absolute paths as the form correct on both sides. That matches this deployment's own measured evidence, and assertion 2 pins it so a later "simplification" back to the unqualified version fails the build.

### Verdict

CI is green (42 success / 9 skipped, zero failures), there are no unresolved threads, no non-thread review findings, and the PR is `CLEAN` against base.

PR #2872, opened 2026-08-17
chore(agents): pin agent-plugins 4.4.0 for the simplification principle

devantler:

### Why

The simplification principle you asked for is merged upstream, but the agents here load from the pinned revision — so until this lands it exists in the marketplace and is absent from every running agent. This is the step that actually delivers it.

### What

Moves the `libraries/agent-plugins` pin to the merged revision (plugin 4.4.0) and syncs the copied desired state so its pinned agent digests match those definitions. No behaviour change beyond the principle itself.

Both roles now prefer the simplest approach that reaches the required outcome over a custom or unproven one — with two boundaries: simplicity is measured against the outcome and never traded for it, and it never justifies removing a control or a measurement.

Part of devantler-tech/agent-plugins#143

### DIFF
--- a/.claude/plugin-consumption/agentic-engineering.desired-state.json
+++ b/.claude/plugin-consumption/agentic-engineering.desired-state.json
@@ -17,7 +17,7 @@
           "executable": true
         }
       ],
-      "entrypointSha256": "ed1f8cf97fbc2d4bdc375a4de048ef2fdcc40ab6612a69ea9e395b7b8e26e5a4",
+      "entrypointSha256": "897ded10f2106b8d166a703cb9473eb25d013ff926580476a4199b7ac163a44b",
       "updatePolicy": "latest-reviewed-default-branch",
       "providerPolicy": "neutral",
       "refreshTiming": "before-starting-each-run",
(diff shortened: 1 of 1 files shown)


devantler:



CI green at `facdaa5` (0 failing). `agent-role-delivery-contract.test.sh` passes against the submodule populated at the new pin — that is the check which hashes the copied desired state against the pinned definitions, so it is what proves the gitlink and the desired-state sync agree rather than merely look consistent.

@coderabbitai review

devantler:

## Readiness at `facdaa5cd544c996cdb421da03689bee9a7a09c8`

**1. Programmatically tested.** Pentad clear: CI green (0 failing), 0 unresolved threads, `CLEAN`, no non-thread findings — the newest CodeRabbit artifact's only finding-shaped section is `🔇 Additional comments (2)`, its excluded informational section. Locally, `agent-role-delivery-contract.test.sh` exits 0 against the submodule populated at the new pin. That is the check that matters here: it hashes the copied desired state against the pinned agent definitions, so a bump carrying a stale copy fails it rather than shipping a silent mismatch.

**2. Reviewed.** `cr@facdaa5` — auto-generated summary updated 12:14:09Z after the request, naming this head, stating *"No actionable comments were generated in the recent review."* Head status `Review completed`. As on the upstream PR, the verdict arrived in the summary with zero review objects at the head.

**3. Tried and evaluated as a user — partially, and the remainder is stated rather than claimed.** What is executable now I ran: the delivery-contract test above, which is the mechanism that would actually break if the gitlink and desired state disagreed. What I have **not** verified is the thing this change exists for — that a dispatched agent loads 4.4.0 and follows the new principle. That is unobservable before merge, because the roles load from the pinned revision and the runtime serves an installed copy that only moves on a refresh plus restart.

So the honest sequence is: merge, then run `plugin-definition-currency.sh`, and treat *its* verdict as what establishes the principle is live. The merge alone does not, and I will not report it as delivered on the strength of the merge.

Promoting and merging.

PR #2911, opened 2026-08-19
fix(docs): clear the dependency advisories blocking every docs update

### LINKED ISSUE #2910: security(docs): an npm audit deadlock has blocked every docs dependency update for a month

## Evidence

The `docs/` site's production dependencies carry **7 published advisories (4 high, 3 moderate)** — in `mermaid`, `nanoid`, `postcss` and `svgo` — and `npm audit` reports **`fix available via npm audit fix`** for every one of them. None is an unfixable transitive pin.

The `docs/package-lock.json` was last updated **2026-07-20**. Since then **no `npm_and_yarn` dependency PR has merged at all** — the last 11 are every one `OPEN` or `CLOSED`-without-merge, while the `submodules` and `github_actions` ecosystems in this same repository merged normally on 08-18 and 08-17. Four are open right now, the oldest created **2026-07-21** (29 days).

Every one of those four fails the same check: **`audit-docs`**.

## The problem — this is a self-sustaining deadlock, not a backlog

`audit-docs` runs `npm audit --omit=dev` over the whole dependency tree. Each Dependabot PR bumps **one** package. So a single-package bump can never make the gate green while any of the other advisories stands:

- every npm PR is permanently blocked →
- the lockfile never advances →
- more advisories accumulate against the frozen tree →
- the gate gets further from green.

It has been closed on itself for a month, and it will not open on its own.

**A second, quieter part of the same defect:** `audit-docs` is gated on `needs.changes.outputs.docs-deps == 'true'`, so it runs *only* when docs dependencies change and **never on an ordinary `main` run**. The check therefore fires exclusively where it blocks and never where it would merely warn — which is why a month of accumulating production advisories left `main` looking entirely green.

Readers of the public devantler.tech docs site, whose delivered assets are built from this tree, and every future docs dependency update, which cannot land until the gate can go green.

## Smallest useful change

Regenerate `docs/package-lock.json` with the available fixes on an **agent-owned** branch. This clears all 7 advisories in one commit, which is the only shape that can satisfy a whole-tree gate — and it unjams the ecosystem, because the next single-package Dependabot PR then rebases onto a tree that already passes.

Dependabot's own PRs are automation-owned and are **not** touched.

## Acceptance criteria

- [ ] `npm audit --omit=dev` exits 0 in `docs/`
- [ ] `audit-docs` green on the PR
- [ ] the docs site still builds
- [ ] a follow-up issue captures the "gate never runs on `main`" blind spot, so accumulation becomes visible before it blocks

**Size:** small — one regenerated lockfile, plus a follow-up issue for the blind spot.

devantler:

### Why

The docs site's dependency updates have been stuck for a month, and the reason is a deadlock rather than a backlog.

The `audit-docs` check audits the **whole** dependency tree, but each Dependabot PR bumps **one** package. So no single-package bump could ever turn the check green while other advisories stood — every docs dependency PR was blocked, the lockfile stopped advancing on 2026-07-20, and further advisories piled up against the frozen tree. Four such PRs are open right now, the oldest from 2026-07-21.

Left alone this does not recover on its own; it gets worse. It also carried real exposure: the published site's dependencies had 4 high and 4 moderate advisories outstanding.

### What

Regenerates the docs lockfile so all outstanding advisories clear in one commit — the only shape that can satisfy a whole-tree check.

Measured in the same Node 22 the CI job uses: the audit fails before this change and reports **0 vulnerabilities** after it. The site still builds — 63 pages, all internal links valid.

**Worth knowing before you merge:** this is a larger movement than a routine patch sweep. `astro` moves 7.0.8 → 7.2.3, which pulls in 11 new transitive packages and drops 2. `package.json` is untouched, so everything stays inside the dependency ranges already declared — but the docs framework itself does move two minors, and that is the main thing to weigh here. The build and link validation both pass on the new tree.

This should also unjam the automation: the next Dependabot PR rebases onto a tree that already passes, so it can go green on its own merits.

Dependabot's own PRs are automation-owned and were not touched.

Fixes #2910

### DIFF
--- a/docs/package-lock.json
+++ b/docs/package-lock.json
@@ -36,29 +36,29 @@
       }
     },
     "node_modules/@astrojs/compiler-binding": {
-      "version": "0.3.1",
-      "resolved": "https://registry.npmjs.org/@astrojs/compiler-binding/-/compiler-binding-0.3.1.tgz",
-      "integrity": "sha512-DaAUj29AIBU2XdJ8uwcab8lW5O2pk9pY8AXkcMw0sw77nVa3oeTYRcO+Dvbbpoexf6ThMc0FMWYCQ/wN1/T7oQ==",
+      "version": "0.3.2",
+      "resolved": "https://registry.npmjs.org/@astrojs/compiler-binding/-/compiler-binding-0.3.2.tgz",
+      "integrity": "sha512-8w/9CWmYrAJJ8N0SY3O43ws2BgxoW6u3QsD8u2mE140lMYAlwh+tlNoUeSBq22wVheFuiBbR212l6ixZ2IIgCQ==",
       "license": "MIT",
       "engines": {
         "node": "^20.19.0 || >=22.12.0"
       },
       "optionalDependencies": {
-        "@astrojs/compiler-binding-darwin-arm64": "0.3.1",
-        "@astrojs/compiler-binding-darwin-x64": "0.3.1",
-        "@astrojs/compiler-binding-linux-arm64-gnu": "0.3.1",
-        "@astrojs/compiler-binding-linux-arm64-musl": "0.3.1",
-        "@astrojs/compiler-binding-linux-x64-gnu": "0.3.1",
-        "@astrojs/compiler-binding-linux-x64-musl": "0.3.1",
-        "@astrojs/compiler-binding-wasm32-wasi": "0.3.1",
-        "@astrojs/compiler-binding-win32-arm64-msvc": "0.3.1",
-        "@astrojs/compiler-binding-win32-x64-msvc": "0.3.1"
+        "@astrojs/compiler-binding-darwin-arm64": "0.3.2",
+        "@astrojs/compiler-binding-darwin-x64": "0.3.2",
+        "@astrojs/compiler-binding-linux-arm64-gnu": "0.3.2",
+        "@astrojs/compiler-binding-linux-arm64-musl": "0.3.2",
+        "@astrojs/compiler-binding-linux-x64-gnu": "0.3.2",
+        "@astrojs/compiler-binding-linux-x64-musl": "0.3.2",
+        "@astrojs/compiler-binding-wasm32-wasi": "0.3.2",
+        "@astrojs/compiler-binding-win32-arm64-msvc": "0.3.2",
+        "@astrojs/compiler-binding-win32-x64-msvc": "0.3.2"
       }
     },
     "node_modules/@astrojs/compiler-binding-darwin-arm64": {
-      "version": "0.3.1",
-      "resolved": "https://registry.npmjs.org/@astrojs/compiler-binding-darwin-arm64/-/compiler-binding-darwin-arm64-0.3.1.tgz",
-      "integrity": "sha512-IEmEF2fUIlTHtpeE/isyEGVOB14cEyh/LZOFYt6wn3jNyVpdC8aR5OZ+RzFUR/f+8ZDM1LaMwZKvoA7eMyJeFw==",
+      "version": "0.3.2",
+      "resolved": "https://registry.npmjs.org/@astrojs/compiler-binding-darwin-arm64/-/compiler-binding-darwin-arm64-0.3.2.tgz",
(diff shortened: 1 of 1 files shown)


devantler:



@coderabbitai review

devantler:

CodeRabbit completed at this head without producing a gate-satisfying artifact: it replied `Action not completed — No files to review`, because the sole changed file is a generated lockfile, which it excludes by default. Zero review objects, zero threads. Recording that as a completed no-gate outcome and advancing to the next lane in priority order rather than re-requesting the same one.




@codex review

devantler:

Codex reported a usage limit for code reviews at this head — a billing limit with no stated retry window, so that lane is not delivering either. Advancing to the third and last lane.

devantler:
@cursor review

devantler (commented):

**Reviewed commit:** `02b65e253540dc376cbff774765b8be6d0c9a056`

### Why the fallback applies — all three lanes tried at this head

- **CodeRabbit** (2026-08-19T06:14:57Z) — completed, but emitted no gate-satisfying artifact: `Action not completed — No files to review`, because the sole changed file is a generated lockfile, which it excludes by default. 0 review objects, 0 threads. Note its head status still reads `Review completed`, which corroborates only that a run happened.
- **Codex** (06:16:21Z) — `You have reached your Codex usage limits for code reviews`. A billing limit with no stated retry window; only the maintainer can lift it.
- **Cursor Bugbot** (06:17:48Z) — check-run `conclusion: neutral` with `output.title: Error`, plus `Bugbot couldn't run - usage limit reached`. Per the lane contract that shape means the review never ran, so it is lane-failure evidence rather than findings. No monthly quota was consumed.

### Findings

**1 finding, raised and addressed — no P0/P1 remaining.**

**nit (accuracy of the PR description, now fixed):** the original body claimed every change was a patch bump inside declared ranges. The first half was wrong. `npm audit fix` moved top-level `astro` **7.0.8 → 7.2.3** (two minors), transitive `cookie` **1.1.1 → 2.0.1** and `neotraverse` **0.6.18 → 1.0.1** (both major), and **added 11 packages while removing 2** (`undici`, `find-process`, `chalk` and its colour chain, `magic-string`, `loglevel`; dropping `@rollup/pluginutils` and its nested `estree-walker`). `package.json` is genuinely untouched, so all of it stays inside the ranges already declared — but describing it as a patch sweep understated what a reviewer is being asked to accept. The body now states the astro movement explicitly.

### What I checked

- **The security claim actually holds:** `npm audit --omit=dev` exits 1 before (8 advisories: 4 high, 4 moderate) and reports `found 0 vulnerabilities` after, run in the same Node 22 the CI job uses.
- **The gate genuinely executed rather than skipping:** the `audit-docs` job ran 06:13:07→06:13:30Z with its `Audit` step `success`. A skipped job would also have shown green, so this was verified rather than assumed.
- **The tree is usable, not merely installable:** the site builds — 63 pages, search index generated, and link validation reports all internal links valid. This is what makes the astro minor bump acceptable rather than a leap of faith.
- **Blast radius is confined to the lockfile:** exactly one file changes; no dependency-policy change rides along.
- **Considered and rejected a narrower fix:** hand-pinning only the four advisory packages would produce a smaller diff but a more fragile tree, and would not survive the next `npm audit fix`. The canonical tool with an unchanged `package.json` is the more durable shape.

### Residual risk the maintainer should weigh

The astro minor bump is broader than the advisories strictly required. It is validated by a passing build and link check, but it is the part of this change most worth a human glance — and it is the reason the honest description mattered.

devantler:

**User evaluation** (readiness condition 3) — exercised at `02b65e253540dc376cbff774765b8be6d0c9a056`, not inferred:

- Ran the real build on the fixed tree (`npm run build`, Node 22): **63 pages built**, Pagefind search index generated from 64 HTML files, and link validation reported **"All internal links are valid."** This is what makes the astro 7.0.8 → 7.2.3 movement acceptable — the site it produces was actually built and checked, not assumed to still work.
- Ran the gate itself both ways: `npm audit --omit=dev` exits **1** on `main`'s lockfile (8 advisories) and **0** on this one.
- Confirmed the CI job genuinely executed rather than skipping — `audit-docs` ran 06:13:07→06:13:30Z with its `Audit` step `success`. A skipped job renders green too, so this needed checking rather than reading off the rollup.

Judged as the change's user: the docs site builds and its links resolve, and the check that has blocked every docs dependency PR for a month now passes.

PR #2913, opened 2026-08-19
ci(security): audit docs dependencies on a cadence, not only when a PR touches them

### LINKED ISSUE #2912: security(ci): the docs dependency audit never runs on main, so advisories accumulate invisibly until they block

## Evidence

`audit-docs` in `.github/workflows/ci.yaml` is gated on `needs.changes.outputs.docs-deps == 'true'`, so it runs **only** when a change touches docs dependencies — and therefore **never on an ordinary `main` run**.

The consequence was measured while fixing #2910: the docs lockfile froze on 2026-07-20 and accumulated advisories for a month while `main` stayed entirely green. The check that would have reported them fired only on Dependabot's PRs, where it did not warn — it *blocked*. By the time anything surfaced, the count had grown past what any single-package bump could clear, which is what produced the deadlock in #2910.

GitHub's own Dependabot alerts *did* see the exposure on the default branch throughout. Nothing in the repository's own CI did.

## The problem

The gate fires exclusively where it blocks and never where it would merely warn. That inverts what you want from a supply-chain check: silent accumulation, then a hard stop at the least convenient moment — on someone else's automated PR, which agents are correctly forbidden from touching.

Skipping the job on unrelated PRs is right — it keeps ordinary PRs fast. The gap is that **no** run observes the tree on a cadence.

## Smallest useful change

Give the audit a scheduled or `push: main` run that **reports** without gating PRs, so accumulation is visible while it is still one or two advisories. Keep the existing PR-time gate exactly as it is.

The two-sided test:
- **Security floor gains** — advisory accumulation against a frozen lockfile becomes observable within a day instead of at an unknown later point, and is no longer discoverable only by blocking an unrelated PR.
- **Everyday path cost** — unchanged. No new required check, no extra wait on any PR; the added run is off the PR path entirely.

## Acceptance criteria

- [ ] the docs dependency audit runs on a cadence independent of whether a PR touched docs deps
- [ ] a newly-published advisory against the current lockfile surfaces without needing a PR
- [ ] no additional required check or added latency on ordinary PRs

**Size:** small — one workflow trigger plus a reporting path.

Found while fixing #2910.

devantler:

### Motivation

Nothing in our own CI ever looks at the docs dependency tree unless a pull request happens to touch it. So advisories build up silently while `main` stays green, and the first anyone hears about them is an automated dependency PR going red — a PR agents are not allowed to touch. That is how the docs lockfile ended up frozen for a month: by the time it surfaced, no single-package bump could clear it.

### Changes

A scheduled workflow now runs the same docs audit once a day, so a new advisory shows up while it is still one or two, and while a normal bump can still fix it.

It deliberately runs the *identical* audit the pull-request gate runs — a gentler one here would pass while the gate still blocked, which would just recreate the deadlock.

It gates nothing: no pull-request trigger, not part of "CI - Required Checks", and no extra wait on anyone's PR. When it does fail, it fails on the default branch, where our existing surveyor already treats that as work to pick up.

The `ci.yaml` edit is a comment only — verified as semantically identical to `main`.

Fixes #2912

### DIFF
--- a/.github/workflows/audit-docs.yaml
+++ b/.github/workflows/audit-docs.yaml
@@ -1,0 +1,52 @@
+name: Audit - Docs Dependencies
+
+# Observes the docs dependency tree on a cadence, independent of whether any pull
+# request touched docs dependencies.
+#
+# The `audit-docs` job in ci.yaml runs only when a change touches docs deps, so on an
+# ordinary day nothing audits the lockfile at all. Advisories therefore accumulate
+# unseen and first surface by blocking an automated dependency pull request — which
+# agents may not touch — at which point no single-package bump can clear them.
+#
+# The audit step below is deliberately IDENTICAL to the `Audit` step of ci.yaml's
+# `audit-docs` job. That is what makes this run predict the gate: a weaker audit here
+# would pass while the gate still blocks, recreating exactly the deadlock this exists
+# to prevent. Keep the two in lockstep.
+#
+# This workflow gates nothing. It has no pull_request trigger, is absent from the
+# "CI - Required Checks" aggregation, and adds no latency to any pull request.
+
+on:
+  schedule:
+    - cron: "23 5 * * *"
+  workflow_dispatch:
+
+concurrency:
+  group: "audit-docs"
+  cancel-in-progress: false
+
+permissions: {}
+
+jobs:
+  audit-docs:
+    name: "Audit docs dependencies"
+    runs-on: ubuntu-latest
+    permissions:
+      contents: read
+    steps:
+      - name: Checkout
+        uses: actions/checkout@3d3c42e5aac5ba805825da76410c181273ba90b1 # v7.0.1
+        with:
+          persist-credentials: false
+      - name: Setup Node.js
+        uses: actions/setup-node@820762786026740c76f36085b0efc47a31fe5020 # v7.0.0
+        with:
+          node-version: "22"
+          cache: "npm"
+          cache-dependency-path: docs/package-lock.json
+      - name: Install dependencies
+        working-directory: docs
+        run: npm ci
+      - name: Audit
+        working-directory: docs
+        run: npm audit --omit=dev
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -418,6 +418,9 @@ jobs:
         working-directory: docs
         run: npm run build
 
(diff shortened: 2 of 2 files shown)


devantler:

@coderabbitai review

devantler:

## Readiness record — head `e6d91f54`

**1. Programmatically tested**
- CI: 59 check-runs, **0 failures**, 0 running. `mergeStateStatus: CLEAN`.
- `actionlint v1.7.9` clean on the new workflow. **Ablated** to prove the check bites: breaking the cron to `not-a-cron` makes it exit 1 with `invalid CRON format`, so the clean result is meaningful rather than vacuous.
- The `ci.yaml` edit is **provably comment-only** — `yq -o=json` of the whole file is byte-identical to `origin/main`, and `jobs.status.needs` is unchanged, so no required check was added.
- The scheduled Audit step is **identical to the gate's** (`docs | npm audit --omit=dev`). **Ablated**: weakening it to `--audit-level=high` is detected. This is the property that matters — a gentler scheduled audit would pass while the gate still blocked, recreating the deadlock.
- Action pins are identical to the gate's (no transposed SHA); zero `${{ }}` interpolation in any `run:` block; no write permissions, no secrets, `persist-credentials: false`.

**2. Reviewed** — CodeRabbit green at this head: head status `Review completed`, summary comment at 10:13:45Z (after the 10:11:57Z request) naming `e6d91f54`, **0 inline findings**, and the only collapsed section is `🔇 Additional comments (2)` (the excluded informational class), both LGTM. Pre-merge checks 5/5 passed.

**3. Tried and evaluated as a user** — stated precisely, because half of this is genuinely not exercisable before merge:
- The **audit step itself is already proven in real CI**: the byte-identical step ran on #2911 at 06:13Z today with `conclusion=success`. That is the enacting path, observed rather than reasoned about.
- The **trigger wiring cannot be exercised pre-merge** — GitHub only runs `schedule`/`workflow_dispatch` from the default branch, so no push to this branch can start it. `actionlint` validates the trigger block statically, and the workflow is correctly absent from the repo's registered-workflow list until it lands.
- I verified the **surfacing claim** rather than asserting it: the plugin's `classify-default-branch-ci-runs.sh` lists `schedule` in its default-branch event set, so a failing scheduled run becomes actionable work rather than a red box nobody reads.

**Committed follow-up:** once merged I will `workflow_dispatch` this workflow and confirm it actually runs and reports, which is the half that cannot be observed from here.

PR #2915, opened 2026-08-19
fix(telemetry): state extraction failure instead of reporting zero busy-waiting

### LINKED ISSUE #2914: Telemetry: a broken extractor jq reports zero busy-waiting instead of a broken instrument

### Why

Every number in the telemetry miner's efficiency section is derived from one extraction pass. When the embedded jq program in that pass has a syntax error, jq compiles nothing, prints its message to a suppressed stderr, and produces no output for **every** file at once. The section then reports `0` busy-wait calls and the script exits successfully.

`0` is exactly what a genuinely improved window looks like. So a one-character typo in the instrument reads as "the agent stopped busy-waiting" — a false improvement, in the one direction that would produce a wrong `WORKING` verdict on the efficiency hypotheses the ledger tracks.

This is not hypothetical. It happened during the 2026-08-19 improver run: a `?//` typo (jq parses `?//` as the alternative-destructuring operator) made a mining pass report **0 tool errors across 182 sessions**, when one file alone held 187. The zero was nearly believed.

The script already guards the neighbouring case — per-line shape aborts — and its own comment says the failure is "SILENT AND ALWAYS DOWNWARD ... exactly the wrong direction for instruments the agent uses to decide whether its own fixes worked". One call site checks the extraction status for this reason; the efficiency call site does not.

### What

Record extraction failures during the efficiency walk and state them, so a broken instrument is distinguishable from a quiet window. Where every file failed, say plainly that the numbers must not be recorded as a measurement.

### Acceptance criteria

- A healthy extractor raises no canary and still reports its count (a canary that always fires guards nothing).
- An extractor broken by a jq program error reports the failure, and the run says the numbers are not a measurement.
- Both states covered by an ablated test wired into CI.

devantler:

### Why

The telemetry miner is how this agent decides whether its own fixes worked. Its efficiency section — every busy-wait number in it — comes from a single extraction pass. If the jq program inside that pass has a syntax error, jq compiles nothing, its message goes to a suppressed stderr, and **every** file yields nothing. The section then reported `0` busy-wait calls and exited successfully.

`0` is indistinguishable from a genuinely quiet window, and it points the way that looks like success. So a one-character typo in the instrument read as "the agent stopped busy-waiting" — which is how a broken measurement becomes a wrong verdict on an open hypothesis.

It happened during this run: a `?//` typo made a mining pass report 0 tool errors across 182 sessions when one file alone held 187.

### What

The efficiency walk now records extraction failures and states them. When every file failed, it says plainly that the numbers are not a measurement rather than printing zeros. A healthy run is unchanged and silent.

Fixes #2914

### DIFF
--- a/.claude/scripts/agent-telemetry.sh
+++ b/.claude/scripts/agent-telemetry.sh
@@ -250,6 +250,12 @@ PROVTMP=$(mktemp "${TMPDIR:-/tmp}/.agtel_prov.XXXXXXXX") || { echo "cannot creat
 CONCTMP=$(mktemp "${TMPDIR:-/tmp}/.agtel_conc.XXXXXXXX") || { echo "cannot create temp file" >&2; exit 3; }
 CREDCONC=$(mktemp "${TMPDIR:-/tmp}/.agtel_credconc.XXXXXXXX") || { echo "cannot create temp file" >&2; exit 3; }
 CREDPROV=$(mktemp "${TMPDIR:-/tmp}/.agtel_credprov.XXXXXXXX") || { echo "cannot create temp file" >&2; exit 3; }
+# Extraction-failure counter for the efficiency walk. A jq PROGRAM error
+# (a typo in the embedded program) makes `tagged_commands_in` emit nothing
+# for EVERY file while exiting non-zero, so the busy-wait metrics below
+# would read 0 and the run would report the agent had stopped busy-waiting.
+# The walk runs in a pipeline subshell, so the count must survive on disk.
+XFTMP=$(mktemp "${TMPDIR:-/tmp}/.agtel_xf.XXXXXXXX") || { echo "cannot create temp file" >&2; exit 3; }
 # Blob-embedded evidence set (#2522): the subset of table values whose
 # occurrences are ALL inside a base64 run. Holds normalised credential values,
 # so it is created with mktemp's private mode and removed by the same traps as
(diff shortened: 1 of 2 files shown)


devantler:


@coderabbitai review

devantler:


@codex review

devantler:


CodeRabbit rate limited and Codex usage-limited at this head; trying Bugbot next.

devantler:
@cursor review

devantler (commented):

**Reviewed commit:** `a9ce3c9e618f059a7e8760cc64726a55c8d7f752`

All three lanes were tried at **this exact head**, each on its own surface, not inherited from a previous run's note:

| Lane | Surface checked | Result |
|---|---|---|
| CodeRabbit | `commits/<head>/status` → `CodeRabbit` context **description** | `Review rate limited` — no retry window stated. (Requested at this head; the head's prior description was the never-reviewed default, so the request was correct.) |
| Codex | issue comments from `chatgpt-codex-connector[bot]` | `You have reached your Codex usage limits for code reviews` |
| Cursor Bugbot | `commits/<head>/check-runs` | `conclusion=neutral` + `output.title=Error` (never-ran shape), with `cursor[bot]`: *Bugbot is counted against Cursor usage … this run hit* the limit. **0** inline comments, consistent with a failed run. |

Codex and Bugbot are **billing/usage** exhausted — maintainer-only to lift. Codex's exhaustion is the same account and credit pool as the currently dead Codex agent lane.

### What I checked

- **Both states of the guard, by ablation.** Healthy extractor → count reported, no canary. Extractor broken by a one-line jq mutation (`(.id? // "")` → `(.id?//"")`, exactly one changed line asserted) → count collapses to 0 **and** the canary fires. Asserting only the canary would have passed even if the mutation had been harmless, so both halves are pinned.
- **False positives at scale.** Ran the real miner over the live 7-day corpus (182 sessions): it produced real numbers (82 timeouts, 627 sleep/poll, 226 primary busy-wait) **and** raised no canary. The zero is meaningful because the run demonstrably ran — checked the output's byte count before believing it.
- **The CI job actually executes.** `test-agent-telemetry` is in the `paths:` filter for both changed files and succeeded on **ubuntu-latest and macos-latest** at this head; a `paths:` grep hit alone would not have shown that.
- **Full suite:** 588 passed, 0 failed.
- **Subshell correctness.** The walk runs in a pipeline subshell, so the failure count is kept on disk rather than in a variable that would not survive; confirmed by the canary firing from inside that loop.

### Findings I raised against myself

**nit — the partial-failure arm is unverified.** When *some* files fail, the code prints an under-count warning instead of the hard one. I could not reach that branch: unreadable and malformed files are filtered out upstream before the loop, so two fixture attempts both left it untested. It is kept as defensive depth, and I am recording it as **unproven rather than tested**.

**P1 (filed, not fixed here) — the fix is narrower than the problem.** Only the efficiency walk is guarded. The reliability, safety and coordination walks use the same suppressed-stderr pattern with no canary, so a broken extractor there still prints zeros — and zero errors / zero leaks / zero races also look like good news. Filed as #2916 rather than widened into this PR, because each remaining walk needs its own decision about what "failed" means on a legitimately quiet window. **Do not read this PR as making the miner fail-closed.**

### Verdict

`Verdict: no P0/P1 findings blocking this change` — the P1 above is a scope limit of the change, filed and stated in the PR rather than left implied, and the nit is disclosed as unproven.

devantler:

### Readiness — all three conditions at `a9ce3c9e618f059a7e8760cc64726a55c8d7f752`

**1. Programmatically tested.** Full suite 588 passed / 0 failed. The guard is pinned in both states by ablation: healthy → count reported and silent; extractor broken by a single-line jq mutation → count collapses to 0 **and** the canary fires. The ablation asserts its own mutation landed (exactly one changed line), so a mis-aimed `sed` fails loudly instead of passing vacuously. CI green: 13 success, 0 failed, and `test-agent-telemetry` confirmed **executed** on both ubuntu-latest and macos-latest.

**2. Reviewed.** CodeRabbit, Codex and Bugbot each tried at this head on their own surface and each verified refusing (rate limit / usage limit / usage limit). Local review round posted at this exact commit, clean.

**3. Tried and evaluated as a user.** I ran the real miner over the live 7-day corpus (182 sessions). It produced genuine numbers — 82 bash timeouts, 627 sleep/poll calls, 226 primary busy-wait — and raised **no** canary, so the guard does not fire on healthy data at scale. I checked the output byte count before trusting that zero, having already been burned once this session by a command that never ran.

Judged as its user: the failure it prevents is one I hit *while building it*. A mining pass of mine reported 0 tool errors across 182 sessions when a single file held 187 — a `?//` typo, silently swallowed. Before this change the deployed miner would have reported that same shape as an improvement in the agent's behaviour.

**Scope, stated plainly:** this guards the efficiency walk only. #2916 covers the remaining walks; this PR does not make the miner fail-closed.

PR #2924, opened 2026-08-19
perf(agents): ratchet the temporary surveyor overlay's byte ceiling

### LINKED ISSUE #2923: perf(agents): surveyor definition load grew 3x unmeasured; the temporary overlay is 2.4x its post-parity size

## Evidence

The `portfolio-surveyor` subagent's **first-turn definition load** has grown ~3× and nothing measures
it. Measured 2026-08-19 over the 7-day session corpus (182 subagent sidechains, 182 extracted, 0
extraction failures):

| Agent | dispatches (7d) | first-turn `cache_creation` median |
|---|---|---|
| `portfolio-surveyor` | **157 (86% of all subagent dispatches)** | **191,397 tok** |
| `general-purpose` | 11 | 158,787 |
| `Explore` | 3 | 29,310 |

`Explore` receives no project instructions; the surveyor does — which is what the ~162K gap is.

**The trend is monotonic, every single day:**

| day | n | median |
|---|---|---|
| 08-13 | 10 | 155,212 |
| 08-14 | 21 | 182,123 |
| 08-15 | 20 | 187,344 |
| 08-16 | 21 | 191,397 |
| 08-17 | 23 | 199,796 |
| 08-18 | 24 | 204,610 |
| 08-19 | 23 | 207,321 |

**+52,109 tokens per dispatch in one week.** And `cache_read` is **0 on 160 of 161** surveyor
dispatches — every dispatch pays the full cache-*write* premium with no reuse, because all subagent
dispatches write the 5-minute cache (`ephemeral_5m` on 182/182, `ephemeral_1h` on 0/182) while the
surveyor is dispatched roughly hourly. At ~157 dispatches/week that is **~30M cache_creation tokens
per week spent before the survey does any work**.

The two inputs, from git history:

| surface | 2026-07-27 | 2026-08-19 | change |
|---|---|---|---|
| `AGENTS.md` | 242,318 B | 387,532 B | **+60%** |
| `.claude/agents/portfolio-surveyor.md` (overlay) | 61,144 B | 148,356 B | **+143%** |

## Problem

`AGENTS.md` calls the surveyor overlay **"explicitly temporary"**, retained "only for
devantler-tech-specific deltas", and states that **generic role logic changes at its owning
upstream**. That routing rule is correct and is not being followed:

- `agent-plugins#78` ("upstream monorepo portfolio-surveyor refinements for digest parity") closed
  **COMPLETED on 2026-07-25** — the parity gate was reached.
- The overlay has grown **2.4× since**, because refinements 4b/4c/4d/4e in
  `.claude/plugin-consumption/agentic-engineering-surveyor-diff.md` (monorepo#2577, #2676,
  #2620/#2713/#2819, #2758) are **generic role logic** that went into the *temporary local* file
  instead of upstream — re-opening the very parity gap #78 had just closed.
- The removal tracker #2363 has not been touched since **2026-08-01**.

Each such append does double damage: it adds to the per-dispatch cost on 86% of all subagent
dispatches, *and* it pushes the overlay's own deletion gate further away.

Nothing in CI, the scorecard, or the run loop measures definition load, which is why a 3× growth in
the deployment's single largest recurring cost went unremarked for a month.

## Hypothesis

The defect is a **missing validation, not a missing rule** — the contract already routes generic role
logic upstream. Making growth *visible at the moment it is authored* converts an invisible append into
a deliberate, reviewed decision.

## Success signal

- **Baseline:** overlay 148,356 B; `AGENTS.md` 387,532 B; surveyor first-turn median 207,321 tok (08-19).
- **Target:** the overlay stops growing without an explicit, justified ceiling raise; new generic
  surveyor refinements land upstream.
- **Guardrail:** the guard must never block mandated work — raising the ceiling in the same PR is
  always available, so the cost is *surfaced*, never vetoed.

## Smallest useful change

A `definition-load-budget-contract.test.sh` ratchet pinning both surfaces' byte ceilings, wired into
CI on the same `AGENTS.md` + self + `ci.yaml` path filter every other contract guard uses, whose
failure message names the fix (upstream it, or raise the ceiling here and say why). Plus a short
*Context & token discipline* note in `AGENTS.md` recording the measured cost.

## Acceptance criteria

- [ ] Guard fails when either surface exceeds its recorded ceiling; passes at current sizes.
- [ ] Failure message names both remedies, including the deliberate raise.
- [ ] Every assertion ablated and verified to fire.
- [ ] Wired into `ci.yaml` (filter + job + aggregate gate).
- [ ] `AGENTS.md` records the measured baseline so the next reader can weigh a raise against it.

**Size:** S–M. Related: #2363 (removal tracker), `agent-plugins#95` (port guards upstream).

devantler:

### Why

The survey subagent runs roughly hourly and is **86% of all subagent dispatches**, so whatever it
loads before doing any work is the largest recurring cost we have. That load has grown **~3×** and
nothing was measuring it — the daily median rose every single day last week, **+52,109 tokens per
dispatch in seven days**. Half of that growth is a file we already decided to **delete**: its parity
gate was reached on 2026-07-25, and it has since grown **+143%** because generic improvements were
appended locally instead of sent upstream, which both charges every dispatch and pushes the file's
own removal further away.

### What

Adds a size ratchet on that temporary file only. Growth now fails CI with a message naming both
ways forward — send it upstream, or raise the limit here and say why. **It never blocks work**:
raising the limit is always available, so the only thing that changes is that the cost becomes a
decision somebody made rather than one nobody saw. The main contract is deliberately left ungated,
because rules legitimately accumulate there and a check that fired on every change would just train
people to click past it.

Fixes #2923

### DIFF
--- a/.claude/scripts/definition-load-budget-contract.test.sh
+++ b/.claude/scripts/definition-load-budget-contract.test.sh
@@ -75,8 +75,8 @@ grep -qF '157 of 182 subagent dispatches' "${constitution}" ||
   fail "AGENTS.md no longer records the dispatch share the cost is multiplied by — a raise could not be weighed without it"
 
 # 3. Both remedies must stay named, so the guard keeps failing WITH the fix rather than just failing.
-grep -qF 'upstream' "${constitution}" ||
-  fail "AGENTS.md no longer names the upstream remedy"
+grep -qF 'goes UPSTREAM unless it is a genuine deployment fact' "${constitution}" ||
+  fail "AGENTS.md no longer names the upstream remedy in the routing sentence. NOTE: do not weaken this to a bare 'upstream' match — that word appears ~49 times in this contract, so the assertion would pass with this whole section deleted."
 grep -qF 'raise the ceiling' "${constitution}" ||
   fail "AGENTS.md no longer names the deliberate-raise remedy, leaving the ceiling looking like a hard veto"
 
--- a/.claude/scripts/definition-load-budget-contract.test.sh
+++ b/.claude/scripts/definition-load-budget-contract.test.sh
@@ -1,0 +1,99 @@
+#!/usr/bin/env bash
+#
+# Ratchets the byte size of the TEMPORARY `portfolio-surveyor` compatibility overlay.
+#
+# Why this needs a guard. The overlay is loaded on every `portfolio-surveyor` dispatch, and that
+# subagent is 157 of 182 subagent dispatches (86%) in the 7-day corpus measured 2026-08-19. Its
+# first-turn `cache_creation` median is 191,397 tokens against `Explore`'s 29,310 — `Explore`
+# receives no project instructions and the surveyor does, so the gap is AGENTS.md plus this overlay.
+# `cache_read` was 0 on 160 of 161 of those dispatches: every subagent writes the 5-minute cache
+# while the surveyor is dispatched roughly hourly, so reuse is structurally zero and every dispatch
+# pays the write premium in full. Daily medians rose monotonically 155,212 -> 207,321 across
+# 08-13..08-19 — +52,109 tokens per dispatch in a single week — and nothing measured it.
+#
+# Why the OVERLAY specifically, and not AGENTS.md. This file is declared temporary: the contract
+# retains it only until digest parity and allows it to carry only its named deployment/provider
+# delta, with generic role logic changing at its owning upstream. That parity gate was reached —
+# agent-plugins#78 closed COMPLETED on 2026-07-25 — and the overlay then grew 61,144 B -> 148,356 B
+# (+143%) because generic refinements were appended here instead of upstreamed, re-opening the gap
+# #78 had just closed. Growth in a file whose declared destination is deletion is always worth a
+# deliberate decision. AGENTS.md is deliberately NOT gated: rules legitimately accrete there, and a
+# ratchet firing on every definition PR — safety fixes included — would train the raise into a
+# reflex and destroy the signal this guard exists to produce.
+#
+# This guard never vetoes mandated work. Raising the ceiling in the same pull request is always
+# available; the ceiling's only job is to make the cost a decision somebody made rather than one
+# nobody saw. Assertion 5 is what keeps a raise honest: the ceiling here and the figure quoted in
+# AGENTS.md must agree, so a raise cannot land without updating the evidence the next reader weighs
+# it against.
+
+set -euo pipefail
+
+repo_root="$(cd "$(dirname "${BASH_SOURCE[0]}")/../.." && pwd)"
+constitution="${repo_root}/AGENTS.md"
+overlay="${repo_root}/.claude/agents/portfolio-surveyor.md"
+
+# The recorded high-water mark, in bytes. Raise this ONLY together with the figure quoted in
+# AGENTS.md's *Context & token discipline* section (assertion 5 enforces that pairing).
+CEILING_OVERLAY_BYTES=148356
+
+fail() {
+  echo "definition-load budget contract: FAIL — $*" >&2
+  exit 1
+}
+
+[ -r "${constitution}" ] || fail "cannot read ${constitution}"
+[ -r "${overlay}" ] || fail "cannot read ${overlay} — the anchor moved, so every assertion below would be vacuous"
+
+overlay_bytes="$(wc -c < "${overlay}" | tr -d ' ')"
+
+# 0. Vacuity guard. A truncated or emptied overlay would sail under the ceiling and pass, which
+# would make this guard report health precisely when the definition had been destroyed.
+[ "${overlay_bytes}" -gt 10000 ] ||
+  fail "overlay is only ${overlay_bytes} B — implausibly small, so the ceiling check below would pass vacuously. Verify the file is intact."
+
+# 1. THE RATCHET. This is the regression itself.
+[ "${overlay_bytes}" -le "${CEILING_OVERLAY_BYTES}" ] ||
+  fail "$(printf '%s' "\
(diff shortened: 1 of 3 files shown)


devantler:

@coderabbitai review

devantler:

CodeRabbit is rate limited at this head (commit-status description `Review rate limited`, plus the rate-limit marker on its summary comment), so no gate-satisfying artifact is coming from that lane. Advancing to the next provider in priority order.

@codex review

devantler:

Codex reports `You have reached your Codex usage limits for code reviews` — a provider quota refusal, not a code verdict. Advancing to the third lane.

devantler:
@cursor review

devantler:

@coderabbitai review

devantler:

CodeRabbit refused again at this head (`Review rate limited`). Advancing.

@codex review

devantler:

Codex refused again at this head (`You have reached your Codex usage limits for code reviews`) — provider quota, not a code verdict. Trying the third lane.

devantler:
@cursor review

devantler (commented):

**Reviewed commit:** `9ed5c43657d2a0039d7fdd7c3ef84e65ac626ef6`

All three external lanes were requested **at this exact head** and each refused on its own surface —
these are provider quota states, not code verdicts, so none is evidence about this change:

| Lane | Surface checked | Result at this head |
|---|---|---|
| CodeRabbit | head commit-status `description` | `Review rate limited` |
| Codex | `chatgpt-codex-connector[bot]` issue comment | `You have reached your Codex usage limits for code reviews` |
| Cursor Bugbot | check-run at head (app slug `cursor`) | `conclusion: neutral` + `output.title: Error` — `Bugbot couldn't run - usage limit reached` (admin must raise the Cursor spend limit; **no retry window stated**) |

### What I checked

- **The ratchet actually ratchets.** Overlay at 148,356 B against a 148,356 B ceiling; +200 B fails.
- **The guard cannot pass vacuously.** Integrity is asserted by the `name: portfolio-surveyor`
  frontmatter rather than a size floor, so a truncated or clobbered file fails at any size.
- **The guard does not fight its own goal.** The overlay's declared destination is deletion, so
  absence passes explicitly and says to retire the guard, and a shrink passes (verified at 4 KB).
- **Prose and enforcement cannot drift.** Assertion 5 binds the ceiling constant to the byte figure
  quoted in `AGENTS.md`, so a raise cannot land without updating the evidence beside it.
- **CI wiring is real, not nominal.** The path filter includes the overlay itself — a PR growing it
  without touching `AGENTS.md` is the target case and would otherwise skip the job — and the job is
  confirmed `SUCCESS` at this head rather than skipped.
- **Arithmetic in the shipped prose** re-derived independently: 157/182 = 86.3%; 148,356/61,144 =
  +143%; 387,532/242,318 = +60%; 207,321 − 155,212 = 52,109.
- **12 assertions ablated**, each verified to fail on *its own* message, plus three positive controls
  (pristine, deleted, shrunk).

### Findings raised during this round — both P1, both already fixed in this head

1. **P1 — near-vacuous assertion.** The upstream-remedy check matched a bare `upstream`, which occurs
   **49 times** in `AGENTS.md`; the entire section could have been deleted with the assertion still
   green. The original ablation concealed this by replacing all 49 occurrences globally — an edit no
   one would make. Fixed by anchoring to the routing sentence and adding a whole-section-deletion
   ablation.
2. **P1 — the guard failed on success.** A deleted overlay hit the anchor check and failed, and the
   10,000-byte floor failed any successful shrink, so completing the upstreaming this guard exists to
   encourage would have broken CI. Fixed as described above.

A third defect was found in the guard by ablation A5b: under `set -euo pipefail` the conservation
extraction died **silently** before its own error could report — a guard unable to announce its own
failure. Fixed with `|| true` and a comment recording why it is load-bearing.

**Verdict: no P0/P1 findings** at `9ed5c43657d2a0039d7fdd7c3ef84e65ac626ef6`.

⚠️ Stated plainly for the record: this verdict is clean *after* the two P1 fixes above, not clean on
first inspection. The round is a fallback for three simultaneously-exhausted providers, and it is
weaker than an independent review — it is my own diff.

devantler:

**Readiness — tried and evaluated as a user, at `9ed5c43657d2a0039d7fdd7c3ef84e65ac626ef6`.**

This change has an exercisable runtime surface (a CI guard), so I ran it rather than reasoning about it.

**Exercised, and what I observed:**
- Ran the guard against the real repository: `PASS — overlay 148356 B / ceiling 148356 B; contract evidence and both remedies intact`.
- Ran it against **12 fixture repositories**, each isolating one conjunct. Every failing case fired on *its own* message — I checked the message, not just a non-zero exit, because a control that errors for an unrelated reason passes vacuously.
- Ran the three **positive** controls, which are the ones that matter for a ratchet: pristine passes; a **deleted** overlay passes and tells the reader to retire the guard; a **4 KB shrink** passes. These exist because the first version of this guard failed all three — it would have blocked the very outcome it exists to encourage.
- Read the failure text as its actual audience: someone whose PR just went red. It names the file, the measured cost, and both remedies including the deliberate raise, so the next step is in the message rather than in a spec.
- Confirmed CI **executed** the job at this head (`Test definition load budget contract = SUCCESS`) rather than skipping it — an all-green PR proves nothing if the guard never ran.

**Judgement as its user:** the guard does the one thing it claims and stops there. It cannot veto mandated work (raising the ceiling is always available and the message says so), it cannot drift from its own evidence (assertion 5 binds the constant to the documented figure), and it cannot be satisfied by a stub. The honest limitation: it enforces the *ceiling*, not the *routing rule* — nothing stops a legitimate deployment-fact append, which is why H48 scores overlay bytes rather than treating a raised ceiling as failure.

PR #2934, opened 2026-08-20
chore(deps): roll agent-plugins to 4.4.1

devantler:

### Why

The agents in this deployment load their definitions from the pinned `agent-plugins` revision. That pin still points at 4.4.0, whose read-only forge guard refuses several of the read commands the portfolio surveyor is instructed to run — so the guard and the instructions disagree, and the guard wins.

The fix for that merged upstream this morning. Until the pin moves, this deployment keeps loading the version with the problem.

### What

Rolls the `libraries/agent-plugins` pin from 4.4.0 to 4.4.1, completing the rollout of `devantler-tech/agent-plugins#144`.

Verified at the new pin: the agent-role delivery contract passes, and the definition-load budget is unchanged (overlay exactly at its ceiling, as before).

One operational note: the currency checker compares the *installed* plugin copy against this pin, so it will report drift until the runtime is refreshed and restarted. That is the expected shape of a pin bump, not a regression — the next run's check is what confirms the new definition is live.

Part of devantler-tech/agent-plugins#144


devantler:

CodeRabbit is structurally unavailable on this PR — its head status reads `Review skipped: ignored keyword in the PR title`, an inherited org/UI setting rather than a transient refusal, so re-requesting it cannot succeed. The title is correct Conventional Commits for a gitlink bump and feeds the changelog, so it is not being reworded to evade the filter. Advancing to the next lane per the review order.

`cr:no-gate@1c3c16f144977df24a96ef74fc2ee294f3acac8d` (reason=title-keyword-skip)



@codex review

devantler:

**Readiness — exercised, not just built.**

The payload of this bump is the forge read-only guard (`a0add262` → `d44dc6dc` changes `forge-readonly-guard.sh`, its test, and the version). So the evaluation that matters is whether the new guard still permits every read this deployment's survey actually makes.

Ran at the new pin, in a worktree at this PR's head with the submodule checked out at `d44dc6dc`:

| check | result |
|---|---|
| `agent-role-delivery-contract.test.sh` | pass |
| `definition-load-budget-contract.test.sh` | pass |
| the guard's own upstream suite | **288 passed, 0 failed** |
| our full survey vocabulary vs the new guard | **38/41 — 3 PROMOTE signals** |

**The three PROMOTE signals are the finding, and they are good news.** They are the exact reads that `agent-plugins#144` tracked as denied by a guard defect — pinned as `gap` so they would fail the moment they started being allowed. This bump is what makes them allowed:

```
gh api -X GET "repos/.../activity" -f per_page=100 -f "ref=refs/heads/main"
gh api --method GET "repos/.../activity" -f per_page=100
gh api -X GET search/issues -f q=org:devantler-tech --paginate
```

Verified both directions rather than inferring: all three **deny** at `a0add262` and **allow** at `d44dc6dc`.

**Merge-order gate.** `.claude/scripts/surveyor-forge-vocabulary.test.sh` is not on `main` yet — it arrives with #2931 — so merging this PR now breaks nothing. But whichever of the two merges **second** must promote those three rows from `gap` to `allow`, or the conformance job goes red on `main`. Since this one is ready first, the promotion belongs on #2931 and will be made there before it merges.

PR #2945, opened 2026-08-20
chore(deps): roll agent-plugins to consume the search-filter guard fix

devantler:

### Motivation

The read-only forge guard that gates the surveyor's access refused two query filters the surveyor's own
definition tells it to use — the filter that finds issues a maintainer commented on, and the one that
bounds a merged-PR window. With the guard enforcing, the maintainer-comment sweep and the merged-PR
retrospective would each stop mid-run rather than return a result.

Upstream has fixed both. This picks the fix up.

### What

Rolls the `agent-plugins` pin forward one commit.

Checked rather than assumed: both filters were replayed against the **new** guard and are now permitted,
and the full corpus of survey commands was replayed with no other verdict moving — so this consumes the
fix without quietly changing what else the boundary allows. The agent and skill definitions are
untouched by the upstream commit, so no role behaviour changes here.

Follow-on, noted so it is not a surprise: #2937 introduces two rows that track these exact refusals as
known gaps. They are correct at today's pin and become stale the moment this merges — that PR promotes
them as part of landing.


devantler:



@coderabbitai review

devantler:



CodeRabbit refused this round at `c83c9967be` — head status reads `Review rate limited`, and its reply
is a rate-limit shell rather than a review. Advancing to the next lane in priority order.

Worth recording: this head's status previously read `Review skipped: ignored keyword in the PR title`,
so the org-level title filter suppresses only the **automatic** review — the explicit trigger did reach
CodeRabbit, which then hit its quota. A `chore(deps)`-titled PR is therefore still reviewable on request.

devantler:



@codex review

devantler:

**Readiness at `c83c9967be`.**

- **Programmatically tested** — 7 contract tests run against the bumped tree, all green:
  `agent-role-delivery-contract`, `definition-load-budget-contract`, `plugin-definition-currency`,
  `surveyor-forge-vocabulary` (41/41), `portfolio-surveyor`, `work-priority-ladder`,
  `review-provider-loop-contract`. CI: 15 pass / 41 skipped / 0 failing.
- **Reviewed** — Codex green at this exact head (`Reviewed commit: c83c9967be`). CodeRabbit refused the
  round with `Review rate limited` (46-minute window), recorded as `cr:no-gate` above. No findings of any
  shape: 0 review objects, 0 unresolved threads, 0 `## Review finding` comments.
- **Tried and evaluated as a user** — the thing this PR changes is what the guard permits, so I exercised
  the guard itself rather than inferring from the version. Against the **new** guard binary: both
  previously-refused filters (`gh search issues --commenter`, `gh search prs --merged-at`) now return
  ALLOW, and replaying all 41 corpus commands moved **no other verdict** in either direction — so the
  fix is consumed without widening the boundary anywhere else. Verified the upstream commit touches only
  the guard and version manifests, leaving every agent and skill definition byte-identical, so no role
  behaviour changes.

One expected operational note: the installed plugin copy will read `DRIFT` against the new pin until the
runtime refreshes it, which is the normal self-healing path and needs no action.

PR #2966, opened 2026-08-21
fix(agents): drop the surveyor's stale repository-visibility markers

devantler:

### Motivation

The portfolio surveyor's own repo list claimed three products were private. All three are public, and have been for some time. That label is not decorative — it feeds the decision about which issues may be added to the shared public project board, where a genuinely private repo's issue must never appear.

The main engineering guide already refuses to record this kind of label anywhere, for exactly this reason: it goes out of date on its own, without anyone touching the file, while still reading as authoritative.

### Changes

Remove the three stale labels. The surveyor already looks this up live wherever it actually matters, so nothing needed replacing — and removing them is better than correcting them, since a corrected label goes stale the same way next time.

Deliberately no explanatory note added in their place: this file is read on almost every background survey, and the rule it would restate is already written down in the main guide.

### DIFF
--- a/.claude/agents/portfolio-surveyor.md
+++ b/.claude/agents/portfolio-surveyor.md
@@ -1315,7 +1315,7 @@ intentional tombstone, never drift.) The list:
 `ksail`, `platform`, `monorepo`, `.github`, `go-template`, `dotnet-template`,
 `platform-tenant-template`, `platform-template`, `actions`, `homebrew-tap`, `agent-skills`,
 `agent-plugins`, `provider-upjet-unifi`, `kyverno-policies`, `maintenance`, `fleet-gitops`, `aws`,
-`world-at-ruin`, `wedding-app` (private), `ascoachingogvaner` (private), `unifi` (private),
+`world-at-ruin`, `wedding-app`, `ascoachingogvaner`, `unifi`,
 `doggy-countdown`.
 Archived repos (currently `reusable-workflows`, `data-product`) are read-only: skip them entirely —
 no CI-red pass, no actionable signal (their stale bot PRs are unmergeable by design).


devantler:


CI is fully green at `bfe20907` — 56 checks, 0 failing, 0 pending — and all 14 `.claude/scripts/*contract*.test.sh` suites pass locally, including `definition-load-budget-contract.test.sh` (overlay 148326 B / ceiling 148356 B; this change is net −30 B).

@coderabbitai review

devantler:

### Exact-head readiness at `bfe20907e69f929b7ca38d30eef32848f8ab82f6`

- **Programmatically tested.** All 14 `.claude/scripts/*contract*.test.sh` suites pass locally, including `agent-role-delivery-contract.test.sh` (after initialising `libraries/agent-plugins`, which is what it was failing on — an environment condition, not this change). CI: **56 checks, 0 failing, 0 pending**. `definition-load-budget-contract.test.sh` PASS at overlay 148326 B / ceiling 148356 B.
  - That guard **failed on my first attempt** at +489 B, because I had added an explanatory note beside the removal. I dropped the note rather than raise the ceiling: the rule it restated already lives in `AGENTS.md`'s portfolio map, and this overlay is loaded on ~86% of subagent dispatches with no cache reuse. Net **−30 B**.
- **Reviewed.** CodeRabbit at this head — substantive summary comment updated 06:51:53Z, naming `bfe20907`, only collapsed section `🔇 Additional comments (1)` (non-actionable), ancillary checks 5 passed / 0 failed. 0 unresolved threads, 0 non-thread findings.
- **Tried and evaluated as a user.** This change is prose consumed by a subagent, so it has **no exercisable runtime surface** — stating that explicitly, as the readiness rule requires for that class. What I did exercise is the underlying claim, live rather than from the file: `gh api repos/devantler-tech/<repo> --jq .private` returns **false** for `wedding-app`, `ascoachingogvaner` and `unifi`, and **true** for `fleet-gitops` — which is the only private repo in the live org set, and the one the surveyor's board-coverage arithmetic already accounts for (673 public open issues on the board out of 675 enumerated).

So the three markers were not merely stale, they were the opposite of the truth on every one of the three.

PR #2979, opened 2026-08-21
fix(agents): stop the unreachable default branch poisoning the sweep's exit code

### LINKED ISSUE #2489: branch-cleanup.sh always exits non-zero from a per-run worktree, so its exit code carries no signal

## Evidence (844th tick, 2026-07-26)

`.claude/scripts/branch-cleanup.sh . monorepo <manifest> apply claude` exits **3** on every invocation from a per-session worktree:

```
monorepo  ns=claude  local: -0  keep 294 | remote: -0  keep 1  rej 0  cand 7  | FAILED to reach main (on 'claude/surveyor-rd-agent-vs-human-2482')
```

The sweep itself **succeeded** — keep-sets computed, 0 deletions, 7 no-PR candidates correctly refused. The non-zero exit comes solely from the final "return to the default branch" step, which fails with:

```
fatal: 'main' is already used by worktree at /Users/homelab-mac-mini/git-personal/monorepo
```

That is not a fault condition. Git refuses to check out one branch in two worktrees, and the *Execution model* mandates per-run worktrees while the main checkout holds `main`. So the duty and the execution model are in direct conflict, and the failure is **structural and unconditional** — every tick hits it.

## Why it matters

The script has real fail-closed paths that must exit non-zero: a failed `git fetch`, a failed open-PR query ("keep-set cannot be trusted"), a failed manifest write. Those are the guards that stop it deleting the head of an open PR — deleting a remote branch **closes** its PR.

Because the routine exit is also non-zero, a caller checking `!= 0` cannot distinguish "the safety guard fired" from "I could not switch branches". The signal that matters most is buried under noise that fires every single run. This is the same signal-degradation class as #2482 and the `$?`-through-a-pipe trap — a check that is always in the failure state stops being read.

## Smallest useful change

Separate the branch-switch from the sweep result. Either:
- treat "default branch is checked out in another worktree" as **success** (it is the expected state under the execution model, and the per-run worktree is discarded anyway), or
- return a distinct exit code for it, so fail-closed aborts stay distinguishable.

## Acceptance criteria

- [ ] A sweep that computes its keep-sets and deletes nothing exits **0** from a per-run worktree.
- [ ] A failed fetch / open-PR query / manifest write still exits non-zero, and RED-provably so.
- [ ] The two are distinguishable by exit code alone, without parsing stdout.

devantler:

### Motivation

Tidying up spent branches is meant to happen at the end of every run. It has not been happening on
this repository, and the reason is structural rather than a mistake anyone made: each run works in
its own private copy of the repository, while the shared copy holds the main branch — and the tool
we use cannot put the same branch in two places at once.

The clean-up itself worked fine. What failed was the very last step, "put the copy back on the main
branch", which simply cannot succeed from a private copy. The tool counted that as a fault and
reported failure every single time, so its success/failure signal became meaningless — a run that
did everything right looked exactly like a run that had genuinely broken. Nearly a thousand old
branches have accumulated behind that noise.

### What this changes

The tool now recognises the difference between "I could not go back to the main branch because
another copy is using it" — which is the normal, expected situation and is left to the scheduled
clean-up to resolve — and "I could not go back for some other reason", which is still reported as a
real failure. Nothing about which branches get kept or removed changes.

Verified on this repository: the same sweep that previously reported failure now reports success,
with an identical list of branches kept and removed.

Fixes #2489

### DIFF
--- a/.claude/scripts/branch-cleanup.sh
+++ b/.claude/scripts/branch-cleanup.sh
@@ -284,6 +284,27 @@ DEFAULT="${DEFAULT:-main}"
 # detached at its superproject pin, and checking out the default branch there
 # would silently move it off the pin (superproject drift).
 START_BRANCH=$(git branch --show-current 2>/dev/null)
+# Is the default branch checked out in some worktree? Only ever consulted after a
+# return-to-default checkout has already FAILED, and only while this checkout sits
+# on some other branch — so a hit is necessarily a DIFFERENT worktree, and no
+# fragile path comparison is needed (macOS resolves /var and /tmp through symlinks,
+# so comparing worktree paths would be the unreliable half of this test).
+#
+# monorepo#2489: every run works in a per-run worktree while the primary checkout
+# holds the default branch, and git refuses one branch in two worktrees. So the
+# return-to-default checkout can NEVER succeed from a session worktree. That is the
+# documented steady state — the scheduled worktree sweep reaps the session worktree
+# and frees its branch — not a fault, so it must not poison the exit code. A
+# checkout that fails for any OTHER reason still must.
+#
+# Fails closed: a failed enumeration (pipefail is on) returns non-zero, so the
+# caller reports the error rather than silently suppressing it.
+default_checked_out_somewhere() {
+  git worktree list --porcelain 2>/dev/null | awk -v want="refs/heads/$DEFAULT" '
+    /^branch / { if (substr($0, 8) == want) found = 1 }
+    END { exit found ? 0 : 1 }
+  '
+}
 return_to_default() {
   local cur
   cur=$(git branch --show-current 2>/dev/null)
@@ -296,6 +317,8 @@ return_to_default() {
     local now
     now=$(git branch --show-current 2>/dev/null)
     if [ "$now" = "$DEFAULT" ]; then sw="-> $DEFAULT";
+    elif default_checked_out_somewhere; then
+      sw="$DEFAULT held by another worktree — left on '${now:-detached}'"
     else sw="FAILED to reach $DEFAULT (on '${now:-detached}')"; errors=$((errors+1)); fi
   else sw="already on $DEFAULT"; fi
 }
--- a/.claude/scripts/branch-cleanup.test.sh
+++ b/.claude/scripts/branch-cleanup.test.sh
@@ -485,6 +485,65 @@ report "control: an unclaimed branch is still deleted at the same list size (#26
 unset WT_SHIM_STATE WT_SHIM_INJECT
 rm -f "$tmp/bin/git"
 
+
+# --- 8. The default branch held by ANOTHER worktree is not an error (#2489) --
+# Every run works in a per-run worktree, and the primary checkout holds the
+# default branch, so `git checkout <default>` inside that worktree can never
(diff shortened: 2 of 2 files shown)


devantler:

## Self-review before requesting a reviewer

**RED/GREEN.** Before the fix: 56 passed / **2 failed**, and the two failures are exactly the new
guard assertions — reproducing the issue's reported output verbatim,
`rc=3 … FAILED to reach main (on 'claude/session-2489')`. After: **58 passed / 0 failed**, exit 0.

**The controls are what make that non-vacuous.** Three of the five new assertions are controls, and
all three pass in the RED run too, so they pin behaviour that had to survive the change:
a reachable default is still returned to, the primary checkout really does move back, and — the
important one — a checkout that fails for a reason *other* than another worktree holding the branch
is **still** an error. That last arm is what stops this from being a blanket suppression of the
failure path.

**Exercised on this repository.** Same real worktree, same sweep, both versions: before → exit **3**,
`FAILED to reach main`; after → exit **0**, `main held by another worktree`. Keep/candidate counts
identical (`keep 978 | remote keep 4 rej 0 cand 31`) and the manifests compare identical, so the
sweep's decisions are untouched and only the signal changed.

**Verified two claims the comment makes rather than assuming them:** `set -uo pipefail` is in force
(line 40), so a failed worktree enumeration returns non-zero and falls through to the error path;
and the script `cd`s into `<repo_path>` (line 74), so the bare `git worktree list` reads the intended
repository.

**One trade-off I want on the record**, since I found it reviewing my own diff rather than being told
it. If the script runs against the *shared* checkout while some stray linked worktree holds the
default branch, that now reports the benign structural case and exits 0, leaving the shared checkout
on its lane branch without a non-zero exit marking it. The diagnosis still appears in the summary
line, and before this change every invocation exited 3 — so that state was already indistinguishable
from the routine one. I judged preserving one-concern scope better than reintroducing a worktree-path
comparison, which is the part macOS symlink resolution makes unreliable.

Ran the correctness pass over the diff; no other findings.

devantler:



CI is green at this head, including the consuming `Test branch cleanup` job on both runners, and the
diff has had its pre-submission self-review (recorded above).

@coderabbitai review

devantler:




CodeRabbit refused at this head in this round (`Review limit reached`, 20:38:22Z, postdating the
request marker above), so recording `cr:no-gate@3d8d8a68` and advancing one lane rather than
re-spending a lane that is demonstrably not serving this head.

@codex review

devantler:

## Ready — promoting

**Green review at head.** Codex reviewed `3d8d8a685a` — the current head — and found no major
issues (20:44:02Z). CodeRabbit had refused this head earlier in the round with `Review limit
reached`, which is service state rather than a verdict, so the lane advanced as recorded above.

**Hygiene at this head:** required checks green, zero unresolved review threads, zero non-thread
review findings on either lane, and no conflict with base.

**User evaluation** is recorded in the self-review above rather than repeated here: the sweep was run
against this real repository before and after, exit `3` → exit `0`, with `keep 978 | remote keep 4
rej 0 cand 31` and the restore manifests identical between the two runs — so the sweep's decisions
are provably untouched and only the exit signal changed.

PR #2985, opened 2026-08-21
test(agents): verify drift in the second runtime asset

### LINKED ISSUE #2982: bug(agents): the definition-currency gate returns UNKNOWN on every run, masking real drift
> 🤖 Generated by the Agent Improver

### Evidence

`plugin-definition-currency.sh` is the mandated pre-flight gate that answers "is the definition the
runtime LOADED the one this consumer PINNED". Since `4b34144` (the routine `agent-plugins` bump
`3b30928` → `7fcf028`, merged 2026-08-21T21:58Z) it returns **UNKNOWN (exit 2) on every invocation**
on this host, producing no verdict at all:

```
$ .claude/scripts/plugin-definition-currency.sh --runtime claude
awk: newline in string scripts/classify-def... at source line 1
rc=2
```

**Mechanism.** Line 309 passes the runtime-asset allow-list into awk as a newline-separated `-v`
assignment and splits it inside `BEGIN`:

```sh
| awk -F'\t' -v p="$prefix" -v runtime="$runtime_assets" '
    BEGIN { count = split(runtime, paths, "\n") ...
```

macOS `/usr/bin/awk` (BSD awk 20200816) **rejects a newline inside a `-v` value**. Minimal repro,
with its control:

```
$ awk -v r="$(printf 'a\nb')" 'BEGIN{print split(r,p,"\n")}'
awk: newline in string a b... at source line 1     # rc=2
$ awk -v r="a"               'BEGIN{print split(r,p,"\n")}'
1                                                   # rc=0
```

**Why now.** `requiredRuntimeAssets` crossed 1 → 2 entries in exactly that commit, which is the
threshold at which `$runtime_assets` becomes multi-line:

| commit | date | assets |
|---|---|---|
| `a32e0e1` | 2026-08-17 | 1 |
| `4b34144` | 2026-08-21T21:58Z | **2** |

So a mechanical dependency bump silently disabled the gate. Nothing signalled it: the contract
correctly says UNKNOWN "must never halt a run", so every lane continues and simply loses the drift
signal.

### Impact

The UNKNOWN is **masking real drift**. Changing only that one awk invocation (`-v` → `ENVIRON`) and
re-running against identical inputs yields a verdict:

```
DRIFT — 3 finding(s) across 11 pinned loaded file(s).
  DRIFT  agents/portfolio-surveyor.agent.md
  DRIFT  scripts/classify-default-branch-ci-runs.sh
  DRIFT  scripts/forge-readonly-guard.sh
```

Every hourly dispatch and every delegated survey has been running a superseded surveyor definition
and a superseded forge guard while the gate reported nothing. `--runtime codex` fails identically;
`--runtime cursor` fails earlier here on an uninitialised submodule and would hit the same line once
populated.

### Why the test suite did not catch it

`plugin-definition-currency.test.sh` (1482 lines) builds its desired-state fixture with **exactly one**
`requiredRuntimeAssets` entry, so the multi-line path is never exercised. The suite is green while the
production configuration — 2 assets since last night — cannot produce a verdict.

### Expected behaviour

The check produces a real verdict (`0` CURRENT / `1` DRIFT / `2` genuinely UNKNOWN) regardless of how
many runtime assets the desired state declares, on both BSD and GNU awk.

### Acceptance criteria

- [ ] The asset list reaches awk by a newline-safe mechanism (`ENVIRON`), not a `-v` assignment.
- [ ] A RED/GREEN test pins a **≥2-asset** desired state and asserts a verdict is produced — failing
      on the current script, passing on the fix.
- [ ] An `awk: newline in string` diagnostic can never again be reported as a clean or unknown run.
- [ ] Verified on macOS BSD awk (the host's `/usr/bin/awk`).

Size: small.

devantler:

### Why

Merged #2986 already restored the macOS definition-currency verdict and added the two-runtime-asset control. One mutation remained unguarded: the check could return a verdict while silently excluding drift in the second declared runtime asset.

### What

- Rebase the stale draft onto current `main` without retaining its superseded production change.
- Mutate the second runtime asset in the existing two-asset fixture and require exit 1 plus the asset path in the DRIFT verdict.

### Verification

- Mutation control: temporarily skipped comparison of the second asset; the new assertion failed because the check incorrectly reported CURRENT.
- Restored implementation: `plugin-definition-currency.test.sh` passes 105 assertions.
- Bash syntax, ShellCheck at warning severity, and `git diff --check` pass.

Closes #2982

### DIFF
--- a/.claude/scripts/agent-role-delivery-contract.test.sh
+++ b/.claude/scripts/agent-role-delivery-contract.test.sh
@@ -428,6 +428,10 @@ assert_prose 'An untouched bot-generated head keeps the repository automation' \
   "consumer does not preserve the existing untouched-bot automation path"
 assert_prose 'Any agent-authored adaptation commit restores the ordinary current-head semantic-review gate' \
   "consumer lets agent adaptations bypass semantic review"
+assert_prose 'convert the PR to draft before the first adaptation push' \
+  "consumer lacks a durable draft fence for dependency-PR adaptations"
+assert_prose 'Draft state is the durable fence' \
+  "consumer trusts reversible auto-merge disarming as the adaptation fence"
 assert_prose 'never select, triage-as-work, edit, or close an issue authored by one of those exact identities' \
   "consumer no longer protects dependency-automation control issues"
 refute_prose 'One bot PR being red, stale, conflicting or review-less remains none of our business' \
--- a/.claude/scripts/plugin-definition-currency.sh
+++ b/.claude/scripts/plugin-definition-currency.sh
@@ -306,9 +306,9 @@ fi
 # behaviour still does not fire. Runtime assets are an explicit allow-list, never all of scripts/.
 reviewed="$(printf '%s
 ' "$tree" \
-  | awk -F'	' -v p="$prefix" -v runtime="$runtime_assets" '
+  | RUNTIME_ASSETS="$runtime_assets" awk -F'	' -v p="$prefix" '
       BEGIN {
-        count = split(runtime, paths, "\n")
+        count = split(ENVIRON["RUNTIME_ASSETS"], paths, "\n")
         for (i = 1; i <= count; i++) required[paths[i]] = 1
       }
       substr($3,1,1)=="\"" { print "QUOTED" ORS; next }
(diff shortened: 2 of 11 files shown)


devantler:
> 🤖 Generated by the Agent Improver



@coderabbitai review

devantler:



@coderabbitai review

devantler:

Superseded by #2987. The attempted non-force history salvage made GitHub present eleven already-merged files instead of the intended test-only boundary. The replacement starts from current `main`, retains the two validated contract tests, and avoids force-pushing this shared branch.

